PR #137, opened 2021-07-06
Improve add item layout

### LINKED ISSUE #56: Improve Add Item Layout
Proposition:

newitemlayout

- "New Item" Button at the right-hand side of the table 
- Probably text edition should happen in a next step, as it might take too much space in this modal.
- There might be a lot of blank space on the right-hand side of the panel, but it will let space for any complementary information. At some point, options such as visiblity, tag, share could also be set there.

abdallah75:
close #56


src/langs/fr.json
```diff
@@ -88,6 +88,12 @@
     "Remove from Favorites": "Retirer des favoris",
     "Add to Favorites": "Ajouter aux favoris",
     "Size": "Taille",
-    "Manage Access": "Gérer les accès"
+    "Manage Access": "Gérer les accès",
+    "New Item": "Nouvel Elément",
+    "Create a Folder": "Créer un Dossier",
+    "Upload a File": "Créer un Fichier",
+    "Create a Link": "Créer un Liens",
```
pyphilia, line 95: `Créer un Lien` ;)

pyphilia (changes requested):
Really good 🔥 Just a minor typo to correct.

pyphilia (changes requested):
Nice ! I really love it. Some suggestions:

- What about switching the "New Item" button and the search bar? I think it would look better, and perhaps at some point the search bar will be less visible so everything would fall nicely in place.
- What do you think of this `text-align: right`? I feel that it is less awkward. (I think it will require some more tweaks than just text-align:right)
Screenshot 2021-07-09 at 15 16 58

src/components/main/TableToolbar.js
```diff
@@ -47,15 +47,18 @@ const TableToolbar = (props) => {
...
+          >
+            {tableTitle}
+          </Typography>
+          {itemSearchInput}
           <NewItemButton fontSize="small" />
-        </Typography>
+        </>
       )}
 
       {numSelected > 0 ? (
```
pyphilia, line 64: Would `numSelected > 0 && ...` work?

src/components/main/ItemTypeTabs.js
```diff
@@ -0,0 +1,96 @@
...
+      orientation="vertical"
+      value={value}
+      onChange={handleChange}
+      className={classes.tabs}
+    >
+      <Tab
+        id={CREATE_ITEM_FOLDER_ID}
+        value={ITEM_TYPES.FOLDER}
+        label={t('Folder')}
+        icon={<FolderIcon className={classes.icon} />}
```
pyphilia, line 57: Could you use `ItemIcon` for icons?

src/components/main/ItemTypeTabs.js
```diff
@@ -0,0 +1,96 @@
...
+    borderRight: `1px solid ${theme.palette.divider}`,
+    minWidth: 'fit-content',
+  },
+  wrapper: {
+    flexDirection: 'row',
+    justifyContent: 'flex-end',
+  },
+  icon: {
+    marginBottom: '0 !important',
+    marginRight: 5,
```
pyphilia, line 30: `theme.spacing`?

pyphilia (approved):
I added some minor comments, you can rebase and apply these requested changes^^

PS: notice `showSaveButton={false}` added to DocumentForm > TextEditor.

Issue #55, opened 2021-04-15
Improve Item Grid Layout


pyphilia:
Ideas:
- [x] Search
- [ ] Filter
- [x] Pagination
- [ ] Order

https://material-ui.com/components/grid-list/

PR #139, opened 2021-07-07
#55 Add pagination to grid view

dialexo:
This PR implements pagination for the grid view, with selectable items per page count as a multiple of 12 (since rows are of either 1, 2, 3, 4 or 6 items in the Grid system):

Peek 2021-07-07 18-38


src/components/main/ItemsGrid.js
```diff
@@ -62,13 +105,52 @@ class ItemsGrid extends Component {
...
         </Grid>
+        <Box
+          display="flex"
+          justifyContent="center"
+          className={classes.paginationControls}
+        >
+          <Box className={classes.itemsPerPageSelect}>
+            <FormControl className={classes.formControl}>
+              <InputLabel id={ITEMS_GRID_ITEMS_PER_PAGE_SELECT_LABEL_ID}>
+                Items per page
```
juancarlosfarah, line 131: Translation

src/components/main/ItemsGrid.js
```diff
@@ -62,13 +105,52 @@ class ItemsGrid extends Component {
...
+              </InputLabel>
+              <Select
+                labelId={ITEMS_GRID_ITEMS_PER_PAGE_SELECT_LABEL_ID}
+                id={ITEMS_GRID_ITEMS_PER_PAGE_SELECT_ID}
+                value={itemsPerPage}
+                onChange={this.handleItemsPerPage}
+                label="Items per page"
```
juancarlosfarah, line 138: Translation

src/components/main/ItemsGrid.js
```diff
@@ -62,13 +105,52 @@ class ItemsGrid extends Component {
...
         <Grid container spacing={1}>
-          {this.renderItems()}
+          {this.renderItems(itemsInPage)}
         </Grid>
+        <Box
+          display="flex"
+          justifyContent="center"
+          className={classes.paginationControls}
+        >
+          <Box className={classes.itemsPerPageSelect}>
```
juancarlosfarah, line 128: I would put this select on the top right. Otherwise it looks a bit weird here. The other option would be to have a proper footer division (e.g. a top border) that indicates that the whole bottom area is a "pagination settings" area. Or maybe change the select style to be contained, so that it looks less like it's just floating in the corner.

juancarlosfarah (changes requested):
Thanks for this! Minor changes only, otherwise a discussion about where to put the select box.

dialexo:
Thanks for the feedback :+1: Missing translations have been added, the select has been moved the right:

image

pyphilia (changes requested):
Really nice pagination^^ Would you like to improve. the cards' layout as well? 

However I noticed a bug when you play around with the search bar + pagination: if you create an element with some title `title` and you search for `title` while being in page 2 or 3, you will have "no results". While searching on page 1 you do have a result. I think it has to do with the number of result which does not reach the page you are.

PR #142, opened 2021-07-08
Implement item flagging

### LINKED ISSUE #119: Implement Item Flagging
Members should be able to flag certain items.

Flags will be stored in a table which links members to the flagged item, and will also contain an optional type/reason for the flag which can be from a predefined set of reasons (might be a separate table)

Members will be able to see all the items that they flagged.
As for admins, they should be able to get some insight regarding flagged items (how many times a certain item was flagged, which members flagged an item and for what reason, most flagged items, ...)

abdallah75:
close #119


pyphilia (commented):
Looks good to me! But before merging this PR, I would wait for the corresponding backend to be merged.

PR #145, opened 2021-07-09
144/document edition

### LINKED ISSUE #144: Enable Document Edition

pyphilia:
close #144


juancarlosfarah (approved):
👍 

Quick question. Why does the `showSaveButton` default to `false`?

pyphilia:
We use the same component in the "add item" modal, which already have an "button", so it's to avoid having twice the same button to save.

pyphilia:
included fix 
close #150

PR #147, opened 2021-07-12
Show shared link in share modal

### LINKED ISSUE #132: Show shared link in Share Modal
The share modal should display a link one can copy to share it with others. This link should be the perform view link. Below it should display the users who can access this item (it will either display a group of users whom have this item's memberships or a sentence: everyone can access this item because it is public)

abdallah75:
close #132


src/components/context/ShareItemModalContext.js
```diff
@@ -45,9 +79,30 @@ const ShareItemModalProvider = ({ children }) => {
   const [open, setOpen] = useState(false);
   const [itemId, setItemId] = useState(null);
 
+  const [linkSource, setLinkSource] = useState(null);
+  const [link, setLink] = useState(null);
+
+  useEffect(() => {
+    if (itemId) {
```
pyphilia, line 86: Perhaps add a case if itemId is not defined.

src/components/context/ShareItemModalContext.js
```diff
@@ -60,13 +115,43 @@ const ShareItemModalProvider = ({ children }) => {
...
-          {/* todo: display perform view link */}
-          {itemId}
+          <div className={classes.shareLinkContainer}>
+            <Link className={classes.shareLink} href={link} target="_blank">
+              {link}
+            </Link>
+            <div>
+              <Select
+                className={classes.selector}
+                value={linkSource}
```
pyphilia, line 138: Maybe `linkType` would be clearer?

pyphilia (changes requested):
Thanks for the PR! Could you provide a screenshot of the visual changes? It would be easier to provide feedback^^

What do you think of not using that grey background and providing white/transparent input? Like the `contained` material UI provides. Let me know what you think.

Maybe it would also make sense to truncate the links to avoid the share modal to resize (when changing from compose to perform), in any case no one is interested in the item id.

src/components/context/ShareItemModalContext.js
```diff
@@ -119,44 +127,53 @@ const ShareItemModalProvider = ({ children }) => {
     navigator.clipboard.writeText(link);
```
pyphilia, line 127: Actually there is a util function in `utils/clipboard.js` `copyToClipboard()` where you can also pass onSuccess and onError. It would be nice to have notification feedbacks!

PR #172, opened 2021-07-20
Show favorite item button in table/grid view

### LINKED ISSUE #151: show that an item is favorite on the table/grid view
This can be done by using a star or heart icon that is filled.

This icon can be on the right where the actions are now:

image

abdallah75:
close #151

### DIFF
--- a/src/components/common/FavoriteButton.js
+++ b/src/components/common/FavoriteButton.js
@@ -14,6 +14,8 @@ const FavoriteButton = ({ item, member }) => {
   const { t } = useTranslation();
   const mutation = useMutation(MUTATION_KEYS.EDIT_MEMBER);
 
+  const isFavorite = isItemFavorite(item, member);
+
   const handleFavorite = () => {
     mutation.mutate({
       id: member.get('id'),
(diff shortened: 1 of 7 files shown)


src/components/common/FavoriteButton.js
```diff
@@ -0,0 +1,73 @@
...
+import { MUTATION_KEYS } from '@graasp/query-client';
+import { FAVORITE_ITEM_BUTTON_CLASS } from '../../config/selectors';
+import { useMutation } from '../../config/queryClient';
+import { isItemFavorite } from '../../utils/item';
+
+const FavoriteButton = ({ item, member }) => {
+  const { t } = useTranslation();
+  const mutation = useMutation(MUTATION_KEYS.EDIT_MEMBER);
+
+  const [isFavorite, setIsFavorite] = useState(isItemFavorite(item, member));
```
pyphilia, line 17: Is it necessary to save `isFavorite` in a state? I think you can just compute the value and use it. But maybe I'm missing something.

pyphilia (approved):
Nice work! I left a minor comment and this is good to go.

PR #177, opened 2021-07-23
105/public items

### LINKED ISSUE #105: Implement Public / Private items
André will create special calls to access public items.

pyphilia:
This PR adds the switch to make an item public. 
While signed out, the request to an item will default to the public link (handled by @graasp/query-client).

close #105

### DIFF
--- a/src/components/item/settings/ItemLoginSetting.js
+++ b/src/components/item/settings/ItemLoginSetting.js
@@ -53,15 +53,17 @@ const ItemLoginSwitch = ({ item }) => {
 
   // update state variables depending on fetch values
   useEffect(() => {
-    const tagValue = getItemLoginTagFromItem({ tags, itemTags });
-    setItemLoginTagValueForItem(tagValue);
-    setIsItemLoginEnabled(Boolean(tagValue));
-    // disable setting if it is set on a parent
-    setIsSwitchDisabled(tagValue && tagValue?.itemPath !== item?.get('path'));
-    setSchema(
-      getItemLoginSchema(item?.get('extra')) ||
-        SETTINGS.ITEM_LOGIN.OPTIONS.USERNAME,
-    );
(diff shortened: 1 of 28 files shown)


src/config/queryClient.js
```diff
@@ -15,6 +15,8 @@ const {
   notifier,
   enableWebsocket: true,
   keepPreviousData: true,
+  // avoid refetching when same data are closely fetched
+  staleTime: 1000,
```
juancarlosfarah, line 19: Is this in seconds? Or in millis?

src/components/item/settings/ItemLoginSetting.js
```diff
@@ -53,15 +53,17 @@ const ItemLoginSwitch = ({ item }) => {
...
-    setIsSwitchDisabled(tagValue && tagValue?.itemPath !== item?.get('path'));
-    setSchema(
-      getItemLoginSchema(item?.get('extra')) ||
-        SETTINGS.ITEM_LOGIN.OPTIONS.USERNAME,
-    );
+    if (tags && itemTags) {
+      const tagValue = getItemLoginTagFromItem({ tags, itemTags });
+      setItemLoginTagValueForItem(tagValue);
+      setIsItemLoginEnabled(Boolean(tagValue));
+      // disable setting if it is set on a parent
```
juancarlosfarah, line 60: Not sure what this means. Does this apply to all ancestors or just the immediate parent?

juancarlosfarah (changes requested):
Looks good! Just some minor comments and questions.

src/components/item/settings/ItemLoginSetting.js
pyphilia: Yes, if the tag exists it is either because the current object or one of its parents has the tag. The check on the path is valid as long as the path is not the current item's.

PR #184, opened 2021-07-29
ci: upgrade yarn to version 2

### LINKED ISSUE #170: Fix yarn crashes on install on Github Actions
Yarn will randomly fail with filesystem errors (e.g. `ENOENT`) on the `install` step, errors seem to stem from cache problems and disk concurrency issues (see https://github.com/yarnpkg/yarn/issues/2629)

Example crash:
```
error https://registry.yarnpkg.com/functional-red-black-tree/-/functional-red-black-tree-1.0.1.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "EEXIST: file already exists, mkdir '/home/runner/.cache/yarn/v6/npm-functional-red-black-tree-1.0.1-1b0ab3bd553b2a0d6399d29c0e3ea0b252078327-integrity/node_modules/functional-red-black-tree'"
```

Recommended action is to use yarn >= v2 (see https://yarnpkg.com/getting-started/migration) as caching should be improved.

Currently Actions runs on `ubuntu-latest` (btw where does it get yarn from?) and `yarn` 1.22.10. Find a way to upgrade yarn workflow.

We may want to use the `setup-node` workflow: https://github.com/actions/setup-node
Example with cypress https://github.com/marketplace/actions/cypress-io#custom-cache-key

Could be applied to other repos as well

pyphilia:
This PR will "very probably" resolve the install error we had in github actions.

close #170


package.json
```diff
@@ -4,8 +4,8 @@
   "private": true,
   "license": "AGPL-3.0-only",
   "dependencies": {
-    "@graasp/query-client": "git://github.com/graasp/graasp-query-client.git#main",
-    "@graasp/ui": "git://github.com/graasp/graasp-ui.git#master",
+    "@graasp/query-client": "git://github.com/graasp/graasp-query-client.git#yarn2",
```
dialexo, line 7: Will we update both main branches for `graasp-query-client` and `graasp-ui` and point to them before merging?

dialexo (approved):

package.json
pyphilia: Yes of course, I'll create the PRs for them right now. They will be merged before this one.

package.json
```diff
@@ -94,8 +94,10 @@
     "cypress-file-upload": "5.0.2",
     "cypress-localstorage-commands": "1.4.5",
     "env-cmd": "10.1.0",
+    "eslint": "^7.31.0",
```
juancarlosfarah, line 97: Can we pin these two repos?

juancarlosfarah (changes requested):
Looks good to me. Just a quick comment regarding pinning.

And what's this file `.yarn/releases/yarn-berry.cjs`? It's different than the one in the other repo.

package.json
pyphilia: It's better not pin them, as it's actually dependent on `react-scripts`'s same dependency. I've tried and it failed.

PR #186, opened 2021-08-02
181/oneLineNameDisplay

### LINKED ISSUE #181: Name is displayed on two lines
ref https://github.com/graasp/graasp-feedback/issues/2

Julien-Torrent:
This prevents longs usernames to be on two lines in the navbar.

closes #181 

With split last name :
image

With a longer last name :
image

### DIFF
--- a/src/components/common/SettingsHeader.js
+++ b/src/components/common/SettingsHeader.js
@@ -117,7 +117,7 @@ const SettingsHeader = () => {
         </Tooltip>
         {username && (
           <Typography variant="subtitle1" className={classes.username}>
-            {truncate(username, { length: USERNAME_MAX_LENGTH })}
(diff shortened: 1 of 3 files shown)


src/components/common/SettingsHeader.js
```diff
@@ -32,7 +32,7 @@ const useStyles = makeStyles((theme) => ({
   },
   username: {
     margin: theme.spacing(0, 2),
-    maxWidth: 100,
+    maxWidth: 120,
```
pyphilia, line 35: It's a detail, but could you factor out this number as `HEADER_USERNAME_MAX_WIDTH` and define it in `constants` right below `USERNAME_MAX_LENGTH`? Then it will be ready to merge!

pyphilia (changes requested):
Good ;) I left a minor comment and we can merge this PR.

PR #191, opened 2021-08-05
refactor: remove obsolete websocket calls, update query-client

dialexo:
Closes #190 by updating the query client and refactoring the websocket calls to use the new API (ws included in hooks)
Automatically, all other behaviours are added by calling the regular hooks.

**todo**: don't forget to update `query-client` first by merging https://github.com/graasp/graasp-query-client/pull/50, and then update `package.json` in this branch with 

```diff
-    "@graasp/query-client": "git://github.com/graasp/graasp-query-client.git#32/ws-included-hooks",
+    "@graasp/query-client": "git://github.com/graasp/graasp-query-client.git#master",
```

PR #195, opened 2021-08-06
fix: discard changes when the modal is closed

Julien-Torrent:
This PR fixes the edition of items.

The editor no longer shows the previous edited value.

closes #191


pyphilia (approved):
Nice catch 👍  Why here tests don't fail? :P

Issue #247, opened 2021-12-03
[Categories] Disable feature if the user cannot edit categories


pyphilia:
The categories should be displayed to the user with read rights only. 

They shouldn't be able to update the data (using the prop `disabled`).

Issue #245, opened 2021-12-02
[Categories] Show categories in alphabetical order and user friendly name


pyphilia:
- `social-science` should be  `Social Science`
- `art` should be displayed before `social-science`
- categories should be translated

PR #253, opened 2021-12-06
fix canEdit for category, add customized tag feature

louisewang1:
image

- fix the canEdit attribute for category selection, when user don't have editting right, the menu will be disabled
- add a input form to set customized tags for the item (in pic above)
- age range => level
- Now options for disciplines are ordered alphabetically, while options for levels maintained the logical order

### DIFF
--- a/src/components/item/sharing/CategorySelection.js
+++ b/src/components/item/sharing/CategorySelection.js
@@ -123,55 +123,51 @@ function CategorySelection({ item, edit }) {
   /* eslint-disable react/jsx-props-no-spreading */
   return (
     <div className={classes.wrapper}>
-      <Typography variant="h6" className={classes.Selection}>
+      <Typography variant="h6" className={classes.selection}>
         {t('Category')}
       </Typography>
-      {edit && ageList && (
-        <>
-          <Typography variant="body1">{t('Age Range')}</Typography>
-          <Autocomplete
-            multiple
-            disableClearable
-            id={SHARE_ITEM_CATEGORY_AGE}
-            value={ageList?.filter((value) => selectedValues.includes(value))}
-            getOptionSelected={(option, value) => option.id === value.id}
-            options={ageList}
-            getOptionLabel={(option) => option.name}
-            onChange={handleChange('age')}
-            renderInput={(params) => (
-              <TextField
-                {...params}
-                variant="outlined"
-                placeholder={t('Please choose from list')}
-              />
-            )}
-          />
-        </>
-      )}
-      {edit && disciplineList && (
-        <>
-          <Typography variant="body1">{t('Discipline')}</Typography>
-          <Autocomplete
-            multiple
-            disableClearable
-            id={SHARE_ITEM_CATEGORY_DISCIPLINE}
-            value={disciplineList?.filter((value) =>
-              selectedValues.includes(value),
-            )}
-            getOptionSelected={(option, value) => option.id === value.id}
-            options={disciplineList}
-            getOptionLabel={(option) => option.name}
-            onChange={handleChange('discipline')}
-            renderInput={(params) => (
-              <TextField
-                {...params}
-                variant="outlined"
-                placeholder={t('Please choose from list')}
-              />
-            )}
-          />
-        </>
-      )}
+      <>
+        <Typography variant="body1">{t('Age Range')}</Typography>
+        <Autocomplete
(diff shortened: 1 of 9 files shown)


louisewang1:
close #252 
close #245 
close #247

src/components/item/sharing/CustomizedTags.js
```diff
@@ -0,0 +1,96 @@
...
+          onChange={handleChange}
+        />
+        <Button
+          type="submit"
+          variant="outlined"
+          color="primary"
+          className={classes.button}
+          endIcon={<SendIcon />}
+        >
+          {t('Submit')}
```
pyphilia, line 89: I would actually use `save` instead, what do you think? (and change the icon or remove it accordingly)

src/components/item/sharing/CustomizedTags.js
```diff
@@ -0,0 +1,96 @@
...
+        {t('Customized Tags')}
+      </Typography>
+      <Typography variant="body1">
+        {t('Please seperate tags by comma. ')}
+        {t('Eg. Tag1,Tag2,Tag3,...,TagN')}
+      </Typography>
+      <form className={classes.container} onSubmit={handleSubmit}>
+        <TextField
+          disabled={!edit}
+          id="customized-tags-input-box"
```
pyphilia, line 74: Unless this refers to an element, remove it.

src/components/item/sharing/CustomizedTags.js
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  return (
+    <>
+      <Typography variant="h6" className={classes.title}>
+        {t('Customized Tags')}
+      </Typography>
+      <Typography variant="body1">
+        {t('Please seperate tags by comma. ')}
+        {t('Eg. Tag1,Tag2,Tag3,...,TagN')}
```
pyphilia, line 69: Very engineer like ;) Maybe let's go for real example tags: `biology`, `moon`, ...

pyphilia (changes requested):
Thanks for this PR! 
I see your branch is still called 237/categories, I think it should be another branch refering to another issue? It's okay for this time, but keep in mind to create new branches for each new issue.
What do you think of `Tags` only instead of `Customized Tags`?

louisewang1:
If I use "Tags", would this be confused with itemTags (like "published", "private")? Besides that I'm good with just "Tags".

src/components/item/sharing/CategorySelection.js
```diff
@@ -62,12 +64,16 @@ function CategorySelection({ item, edit }) {
...
     ?.toArray();
   const disciplineList = categoriesMap
-    ?.get(categoryTypes?.filter((type) => type.name === 'discipline').get(0).id)
-    ?.toArray();
+    ?.get(
+      categoryTypes?.find((type) => type.name === CATEGORY_TYPES.DISCIPLINE)
+        ?.id,
+    )
+    ?.toArray()
+    .sort(compare);
```
pyphilia, line 76: I would rename the sorting function to something more explicit.

src/components/item/sharing/CategorySelection.js
```diff
@@ -123,55 +129,51 @@ function CategorySelection({ item, edit }) {
...
+          value={levelList?.filter((value) => selectedValues.includes(value))}
+          getOptionSelected={(option, value) => option.id === value.id}
+          options={levelList}
+          getOptionLabel={(option) => option.name}
+          onChange={handleChange('level')}
+          renderInput={(params) => (
+            <TextField
+              {...params}
+              variant="outlined"
+              placeholder={t('Please choose from list')}
```
pyphilia, line 151: `Please choose from the list`?

src/components/item/sharing/CustomizedTagsEdit.js
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  return (
+    <>
+      <Typography variant="h6" className={classes.title}>
+        {t('Tags')}
+      </Typography>
+      <Typography variant="body1">
+        {t('Please seperate tags by comma. ')}
+        {t('Eg. English,Biology,Lab,Plants,...,Demo')}
```
pyphilia, line 69: Add spaces between values

pyphilia (changes requested):
Nice PR! I left some comments 🐰

PR #267, opened 2021-12-22
feat: add button to show/hide item in perform

### LINKED ISSUE #107: Implement Show / Hide for items
backend: Use tags -> don't allow if tag is set... but it is against the core permissions.

Another solution would be to hide it only from the front-end.

Julien-Torrent:
close #107


cypress/integration/item/hide/hideItem.spec.js
```diff
@@ -0,0 +1,93 @@
...
+import { HOME_PATH } from '../../../../src/config/paths';
+import {
+  buildItemCard,
+  buildItemsTableRowIdAttribute,
+  HIDDEN_ITEM_BUTTON_CLASS,
+} from '../../../../src/config/selectors';
+import { TABLE_ITEM_RENDER_TIME } from '../../../support/constants';
+import { ITEM_LAYOUT_MODES } from '../../../../src/enums';
+
+const HiddenTagId = 'hiddenTagId';
```
pyphilia, line 11: camelcase.

src/components/common/HideButton.js
```diff
@@ -0,0 +1,63 @@
...
+
+  const isHidden =
+    tags?.filter(({ tagId }) => tagId === HIDDEN_ITEM_TAG_ID).size > 0;
+
+  const handlePin = () => {
+    if (isHidden) {
+      removeTag.mutate({
+        id: item.id,
+        tagId: tags?.filter(({ tagId }) => tagId === HIDDEN_ITEM_TAG_ID).first()
+          .id,
```
pyphilia, line 29: You will need to add `.?` for first and id too. Eventually I guess it's more useful to save the itemTagId instead of just the boolean.

pyphilia (changes requested):
🍈

pyphilia (approved):

PR #294, opened 2022-02-10
refactor: update app item data

### LINKED ISSUE #293: Update App Item props

pyphilia:
close #293

### DIFF
--- a/src/components/main/ItemScreen.js
+++ b/src/components/main/ItemScreen.js
@@ -4,7 +4,7 @@ import { ItemLoginAuthorization } from '@graasp/ui';
 import { useParams } from 'react-router';
 import {
   getMembership,
-  isItemUpdateAllowedForUser,
+  getBestPermissionForMemberFromMemberships,
 } from '../../utils/membership';
 import ErrorAlert from '../common/ErrorAlert';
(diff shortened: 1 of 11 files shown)


swouf (approved):
Ça me semble ok.

src/utils/membership.js
```diff
@@ -18,6 +12,30 @@ export const isItemUpdateAllowedForUser = ({ memberships, memberId }) =>
     ),
   );
 
+export const getBestPermissionForMemberFromMemberships = ({
```
juancarlosfarah, line 15: Maybe just add a description of what is `best`? From what I see here quickly, it's by length?

src/components/main/ItemScreen.js
```diff
@@ -53,10 +54,12 @@ const ItemScreen = () => {
     return <ErrorAlert />;
   }
 
-  const enableEdition = isItemUpdateAllowedForUser({
+  const itemMembership = getBestPermissionForMemberFromMemberships({
```
juancarlosfarah, line 57: Would be good to explain what `best` is?

juancarlosfarah (changes requested):
@pyphilia, looking good! Just think that we should define in the function and where it's used what `best` means in this context.

PR #299, opened 2022-02-24
141/errors trans

### LINKED ISSUE #141: Translate Notification Feedback messages
This might already get advantage of i18n indexing of translations. For example, the keys could be `item.edit.message.failure`, `item.edit.message.success`.

https://www.i18next.com/translation-function/essentials

pyphilia:
close #141


src/components/main/ItemScreen.js
```diff
@@ -75,7 +77,10 @@ const ItemScreen = () => {
 
   return (
     <Main>
-      <ItemMain item={item}>{content}</ItemMain>
+      <UppyContextProvider enable={enableEdition} itemId={itemId}>
```
juancarlosfarah, line 80: Not specifically for this PR, but in English, this should be `enableEditing`.

src/components/file/FileUploader.js
```diff
@@ -48,53 +40,20 @@ const useStyles = makeStyles((theme) => ({
...
-        itemId,
-        onComplete,
-        onFilesAdded,
-        onError,
-        onUpload,
-      }),
-    );
-  };
+  useEffect(() => {
+    uppy?.on('files-added', () => {
```
juancarlosfarah, line 53: Won't `uppy` always have the `on`? Or is this checking if `uppy` is not falsy?

juancarlosfarah (approved):
Approving! Looks very nice. Much more compact. Just a quick question and an unrelated comment on naming.

src/components/file/FileUploader.js
pyphilia: Uppy could be not configured yet.

src/components/main/ItemScreen.js
pyphilia: It's not too much changes, I've applied it 👍

juancarlosfarah (approved):
👍

PR #305, opened 2022-03-16
refactor: use graasp button

### LINKED ISSUE #304: Use Graasp Button from graasp-ui

pyphilia:
This PR updates some properties in TextEditor / DocumentItem corresponding to the changes in graasp-ui.
It also removes all buttons to use the Button in @graasp/ui.

close #304

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@graasp/ui": "github:graasp/graasp-ui.git",
     "@graasp/utils": "github:graasp/graasp-utils.git",
     "@material-ui/core": "4.12.3",
-    "@material-ui/icons": "5.0.0-beta.4",
+    "@material-ui/icons": "5.0.0-beta.5",
     "@material-ui/lab": "4.0.0-alpha.60",
     "@uppy/core": "2.1.6",
     "@uppy/dashboard": "2.1.4",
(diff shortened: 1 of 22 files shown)


alvrba:
I have tested `graasp-compose` and `graasp-ui` and it seems the buttons are working fine. I have two doubts:
- Do we want to change all the buttons to the graasp-ui button? For example, in the NewButtonItem component we are still using the material-ui button directly.
- Graasp-compose is crashing when creating a new item of type `File` or `Import APP`. I see the uppy module has been updated to a new version, and I have tried with the previous one, but it is still giving me the error. I do not know if maybe is my local problem because I need something else (but it is working fine with the Graasp compose in main). Can you check if it is working for you?

pyphilia:
@alvrba Indeed you're right, the app is crashing (I did it quite quickly yesterday). The problems comes from both react-scripts v5 and uppy https://github.com/facebook/create-react-app/pull/12021 I've downgraded react-scripts. Though the whole idea was to remove the security alerts but I guess we will keep them for the time being.

And we want to replace all the buttons yes!

package.json
```diff
@@ -7,47 +7,47 @@
     "@graasp/chatbox": "github:graasp/graasp-chatbox.git",
     "@graasp/query-client": "github:graasp/graasp-query-client.git",
     "@graasp/translations": "github:graasp/graasp-translations.git",
-    "@graasp/ui": "github:graasp/graasp-ui.git",
+    "@graasp/ui": "github:graasp/graasp-ui.git#103/spaceGlitchTextEditor",
```
alvrba, line 10: Do we have to leave this branch or is only for testing?

alvrba (approved):

PR #316, opened 2022-04-01
291/update parent membership

### LINKED ISSUE #291: Updating an item membership from a parent in a child edits the parent's
When adding a permission to an item, this permission propagates to its children.
The current interface allows you to see and edit an item's parent's permission in the child sharing settings. 
It is currently possible to edit the parent's permissions. 

The expected result is that the permission is upgraded in this child (you cannot downgrade a permission in a child). 

Currently, it edits the parent's permission directly instead of creating a new item-membership.

To reproduce:
- set parent item-membership to read
- set child's item-membership to write
- you will see the parent's permission became write -> it should stay read

pyphilia:
close #291

### DIFF
--- a/cypress/integration/memberships/editItemMembership.spec.js
+++ b/cypress/integration/memberships/editItemMembership.spec.js
@@ -37,4 +37,23 @@ describe('Edit Membership', () => {
       expect(body?.permission).to.equal(permission);
     });
   });
+
+  it('edit children membership should create a new membership', () => {
+    cy.setUpApi({ ...ITEMS_WITH_MEMBERSHIPS });
+
+    // go to children item
+    const { id, memberships } = ITEMS_WITH_MEMBERSHIPS.items[1];
+    cy.visit(buildItemPath(id));
+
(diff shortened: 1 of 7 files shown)


cypress/integration/memberships/editItemMembership.spec.js
```diff
@@ -37,4 +37,23 @@ describe('Edit Membership', () => {
...
+
+    // update membership
+    const permission = PERMISSION_LEVELS.ADMIN;
+    const { id: mId } = memberships[1];
+    editItemMembership({ itemId: id, id: mId, permission });
+
+    cy.wait('@shareItem').then(({ request: { url, body } }) => {
+      expect(url).to.contain(id);
+      expect(body?.permission).to.equal(permission);
+      // expect(body?.memberId).to.equal(member.id);
```
spaenleh, line 56: Remove ?

spaenleh (changes requested):
Just a few comments 🐨

PR #323, opened 2022-04-07
feat: add drop down between apps in header

### LINKED ISSUE #176: Add Navigation Button for different apps
Add a button in the nav bar next to the logo. This button opens a dropdown that contains the different graasp apps (compose, perform, explore) to which the user can navigate. We must also navigate to the app deployed within the same environment (production, staging, ...)

pyphilia:
Screenshot 2022-04-07 at 11 10 09

This PR changes the logo and adds the navigation drop down. It can clearly be better, but that's a first version.

close #176


spaenleh (approved):
🔥

PR #326, opened 2022-04-10
318/item publish tab

### LINKED ISSUE #318: New Tab for item publish

louisewang1:
close #318 

Seperate the publish item from sharing tab. Created a new tab for item publishing.


src/components/item/header/ItemHeaderActions.js
```diff
@@ -87,6 +88,7 @@ const ItemHeaderActions = ({ onClickMetadata, onClickChatbox, item }) => {
             </Tooltip>
           )}
           <ShareButton itemId={id} />
+          <PublishButton itemId={id} />
```
pyphilia, line 91: Most probably we want to display this button only when the item is public and the user is an admin.

src/components/item/publish/ItemPublishConfiguration.js
```diff
@@ -128,10 +164,12 @@ const ItemPublishConfiguration = ({
   };
 
   const handleRefresh = () => {
-    queryClient.invalidateQueries(buildItemValidationAndReviewsKey(itemId));
+    queryClient.invalidateQueries(buildItemValidationAndReviewKey(itemId));
   };
 
   const publishItem = () => {
+    const publishedTag = getTagByName(tags, SETTINGS.ITEM_PUBLISHED.name);
+    const publicTag = getTagByName(tags, SETTINGS.ITEM_PUBLIC.name);
```
pyphilia, line 172: So if we handle public in the other dropdown, no need to set the public tag again.

src/components/item/sharing/VisibilitySelect.js
```diff
@@ -254,7 +245,7 @@ const VisibilitySelect = ({ item, edit }) => {
             {t('Pseudonymized')}
           </MenuItem>
           <MenuItem value={SETTINGS.ITEM_PUBLIC.name}>{t('Public')}</MenuItem>
-          <MenuItem value={SETTINGS.ITEM_PUBLISHED.name}>
+          <MenuItem value={SETTINGS.ITEM_PUBLISHED.name} disabled>
```
pyphilia, line 248: Remove.

src/components/item/publish/ItemPublishConfiguration.js
```diff
@@ -250,6 +288,10 @@ const ItemPublishConfiguration = ({
         )}
       </Typography>
       <Button
+        disabled={
+          isDisabled ||
+          itemValidationStatus !== ITEM_VALIDATION_STATUSES.SUCCESS
+        }
         variant="outlined"
         onClick={publishItem}
         color="primary"
```
pyphilia, line 297: If you factor out this button in a new component, you will be able to split the logic of tags / validation.

src/components/item/publish/ItemPublishConfiguration.js
```diff
@@ -87,37 +101,59 @@ const ItemPublishConfiguration = ({
...
-  const validItemValidation = itemValidationData?.filter(
-    (entry) =>
-      new Date(entry.validationUpdatedAt) >= new Date(item?.get('updatedAt')),
-  );
+    useItemValidationAndReview(itemId);
+  // check if validation is still valid
+  const iVId =
+    new Date(itemValidationData?.createdAt) >= new Date(item?.get('updatedAt'))
+      ? itemValidationData?.itemValidationId
+      : undefined;
```
pyphilia, line 109: It's better to set to null or false.

pyphilia (changes requested):
Looking forward this new tab! I've suggested some improvements, let me know if you need more details.

src/components/item/sharing/VisibilitySelect.js
louisewang1: This can cause a problem that when item is published, the selec dropdown menu will be empty

PR #335, opened 2022-04-26
Fix item name Field and App name Field

### LINKED ISSUE #330: Cannot edit app name once you choose an app in the menu

### LINKED ISSUE #332: Can not completely clear an item name (textfield)
When editing an item name, and removing the full string of the name, the old name is re-added to the field.

Expected behaviour:
- ability to completely clear the nam field to change the item's name

Current behaviour:
- once the name field is clear, the old name appears again in the field.
- It is not possible to have an empty field field when submitting but this should be handled on the buttons

Possible fix:
- use `??` instead of `||` to allow for an empty string
- should check that the name is not `''` when wanting to submit and save button should be disabled when name is empty

@pyphilia happy to discuss this issue

spaenleh:
This PR fixes:
- Editing name of item field
- Editing App name field after selecting app in dropdown
- Publish icon size
- Missing span for Tooltip on hidden IconButton for HideButton

Tests now do a `{selectall}{backspace}some text` in the type instead of `{selectall}some text` so that the empty field can be tested too.

closes #332 
closes #330

### DIFF
--- a/cypress/integration/item/edit/editFolder.spec.js
+++ b/cypress/integration/item/edit/editFolder.spec.js
@@ -2,8 +2,9 @@ import { DEFAULT_ITEM_LAYOUT_MODE } from '../../../../src/config/constants';
 import { ITEM_LAYOUT_MODES } from '../../../../src/enums';
 import { buildItemPath, HOME_PATH } from '../../../../src/config/paths';
 import { EDITED_FIELDS, SAMPLE_ITEMS } from '../../../fixtures/items';
-import { EDIT_ITEM_PAUSE } from '../../../support/constants';
+import { EDIT_ITEM_PAUSE, TABLE_ITEM_RENDER_TIME } from '../../../support/constants';
 import { editItem, editCaptionFromViewPage } from './utils';
+import { buildEditButtonId, ITEM_FORM_CONFIRM_BUTTON_ID } from "../../../../src/config/selectors";
 
 describe('Edit Folder', () => {
   describe('List', () => {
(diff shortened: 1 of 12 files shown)


cypress/integration/item/edit/editFolder.spec.js
```diff
@@ -28,6 +29,30 @@ describe('Edit Folder', () => {
       });
     });
 
+    describe('Dumb user', () => {
```
pyphilia, line 32: aha I think you don't need to wrap with the describe

cypress/support/commands/item.js
```diff
@@ -124,7 +127,7 @@ Cypress.Commands.add(
     cy.fillBaseItemModal({ name }, { confirm: false });
 
     cy.get(`#${ITEM_FORM_APP_URL_ID}`).click();
-
+    // todo: can not edit name field in modal
```
pyphilia, line 130: We could add the textfield for the name in the edit modal if you want. Is it referring to that?

src/components/context/EditItemModalContext.js
```diff
@@ -14,6 +14,10 @@ import { ITEM_FORM_CONFIRM_BUTTON_ID } from '../../config/selectors';
 import { ITEM_TYPES } from '../../enums';
 import BaseItemForm from '../item/form/BaseItemForm';
 import DocumentForm from '../item/form/DocumentForm';
+import { isItemValid } from '../../utils/item';
+
+// time to be considered between 2 clicks for a double-click (https://en.wikipedia.org/wiki/Double-click#Speed_and_timing)
+const DOUBLE_CLICK_DELAY_MS = 500;
```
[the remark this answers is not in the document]
pyphilia: Would it be possible to factor this logic in the graasp button (in ui)? (not to do immediately, but it might be a nice to apply this prevention on all our buttons)

pyphilia (changes requested):
Nice fixes 🦛

cypress/support/commands/item.js
spaenleh: I put that there to remind me of adding a test that would try to change the app name. I obviously forgot to add it. Doing so in the next commit.

src/components/context/EditItemModalContext.js
```diff
@@ -44,9 +48,21 @@ const EditItemModalProvider = ({ children }) => {
...
+    // todo: factor out this logic to graasp-ui
+    setTimeout(() => setConfirmButtonDisabled(false), DOUBLE_CLICK_DELAY_MS);
   };
 
   const submit = () => {
+    if (isConfirmButtonDisabled) {
+      return;
+    }
+    if (!isItemValid({ ...item, ...updatedProperties })) {
+      // todo: notify user
```
pyphilia, line 61: Todo?

pyphilia (approved):
I've left a last comment, but it's good to go 🦏

PR #336, opened 2022-04-26
fix: empty email field after invite completes

### LINKED ISSUE #334: Clear email when inviting in share tab

spaenleh:
This PR fixes the issue where the email textfield would keep the email value after inviting a member.

closes #334


pyphilia (approved):
💯

PR #350, opened 2022-05-09
feat: add Collapse item option

alvrba:


pyphilia (approved):
Looks all good to me! Maybe it would be really nice to have tests?

PR #357, opened 2022-05-10
refactor: allow remove and change on cc license

### LINKED ISSUE #347: Allow removing CC License
Finally, we will allow to remove CC License (discussion)

louisewang1:
close #347


src/components/item/publish/CCLicenseSelection.js
```diff
@@ -115,24 +123,42 @@ const CCLicenseSelection = ({ item }) => {
...
+        disabled={!optionValue}
         className={classes.button}
+        buttonName={t('Submit')}
+        dialogContent={t(
+          'Please verify that your item fits the CC License, and do not change to a more restricted option.',
+        )}
         handleSubmit={handleSubmit}
       />
+      {settings?.ccLicenseAdaption !== '' && (
+        <CCLicenseDialog
```
pyphilia, line 145: I don't really think this works well. 
If `settings?.ccLicenseAdaption !== ''` is `true`, then you display both dialog when `open` becomes `true`.
Plus `settings?.ccLicenseAdaption !== ''` means "is my cc license is (already) empty" -> so you don't actually need to remove it? 

I think you need at least to variables to control `open` of each dialog. And have at least a `if/else` close to display only one of the dialog.

src/components/item/publish/CCLicenseSelection.js
```diff
@@ -84,6 +82,16 @@ const CCLicenseSelection = ({ item }) => {
...
   };
 
+  const handleRemove = (event) => {
+    event.preventDefault();
+    updateCCLicense({
+      id: itemId,
+      name: itemName,
+      settings: { ccLicenseAdaption: '' },
+    });
+    setOpen(false);
```
pyphilia, line 92: I see it's actually your CCDialog component which controls the opening (since it has the button). So you should define the open state in CCDialog. instead of closing the dialog in `handleRemove`, do it in CCDialog: 
```js
const onClick = () => {
    handleSubmit()
    setOpen(false)
}
```

src/components/item/publish/CCLicenseSelection.js
```diff
@@ -115,24 +123,42 @@ const CCLicenseSelection = ({ item }) => {
...
+        <CCLicenseDialog
+          open={open}
+          setOpen={setOpen}
+          disabled={!optionValue}
+          className={classes.button}
+          buttonName={t('Remove License')}
+          dialogContent={t(
+            'Are you sure to remove existing CC License? This action is not suggested, and may lead to license inconsistency.',
+          )}
+          handleSubmit={handleRemove}
```
pyphilia, line 154: You didn't provide a screenshot, but I suppose removing the license is a button. Would it make sense to have it in the options instead? Something like `None`?

pyphilia (changes requested):
Thank you for the PR, I've added my comments, we can discuss if something isn't clear!

pyphilia (approved):

PR #382, opened 2022-06-16
356/switch users

### LINKED ISSUE #356: Handle/Switch between main user / light user
Would it be possible to register a normal user as a light user?  -> possibility to switch to another user?
Or a main user would "contain" a light user.

pyphilia:
This PR introduces a new Member Menu with all possible members you can switch to. These are the previous account you connect to. There's plenty of improvement possible: show the item related to the light user, etc.

Screenshot 2022-06-16 at 11 18 47
Screenshot 2022-06-16 at 11 20 25

close #356


spaenleh (approved):
Nice ! Looks good to me ! Let's switch those accounts ! 👤🔄

cypress/integration/header.test.js
```diff
@@ -1,13 +1,68 @@
...
   it('App Navigation', () => {
-    // check navigation and display and interface doesn't crash
     cy.setUpApi();
     cy.visit(HOME_PATH);
+    // check navigation and display and interface doesn't crash
     cy.get(`#${APP_NAVIGATION_DROP_DOWN_ID}`).click();
     cy.wait(3000);
     cy.get(`#${APP_NAVIGATION_DROP_DOWN_ID}`).should('exist');
   });
+  describe('User Menu', () => {
```
dialexo, line 25: style: spacing between `describe` blocks?

cypress/integration/header.test.js
```diff
@@ -1,13 +1,68 @@
...
+      cy.get(`#${HEADER_MEMBER_MENU_BUTTON_ID}`).click();
+      cy.get(`#${HEADER_MEMBER_MENU_SEE_PROFILE_BUTTON_ID}`).click();
+      cy.get(`#${MEMBER_PROFILE_MEMBER_NAME_ID}`).should('be.visible');
+
+      // sign in as another user
+      cy.get(`#${HEADER_MEMBER_MENU_BUTTON_ID}`).click();
+      cy.get(`#${HEADER_MEMBER_MENU_SIGN_IN_BUTTON_ID}`).click();
+      cy.url().should('equal', SIGN_IN_PATH);
+    });
+    it('Sign out', () => {
```
dialexo, line 39: style: spacing between `it` blocks?

cypress/integration/redirection.spec.js
```diff
@@ -0,0 +1,22 @@
...
+  it('Redirection to saved url', () => {
+    const link = 'http://somelink.com';
+    saveUrlForRedirection(link);
+
+    cy.setUpApi();
+    cy.visit(REDIRECT_PATH);
+
+    cy.url().should('contain', link);
+  });
+  it('Redirection to home if no url is saved', () => {
```
dialexo, line 15: style: spacing between `it` blocks?

src/components/common/UserSwitchWrapper.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <>
+      <GraaspUserSwitch
+        ButtonContent={ButtonContent}
+        navigate={navigate}
+        signOut={signOut}
+        currentMember={member}
+        isCurrentMemberLoading={isLoading}
+        isCurrentMemberSuccess={isSuccessUser}
```
dialexo, line 39: I see that all values from `CurrentUserContext` are passed down as props. Is there a reason why `GraaspUserSwitch` is not invoking `useContext` itself? (I guess it is so that GraaspUserSwitch only has UI concerns). Though in which case, would it make sense for GraaspUserSwitch to take the context itself as props? The type could be restricted to the relevant props only from UI side. Not sure what's best 🤔

src/components/context/CurrentUserContext.js
```diff
@@ -17,8 +17,11 @@ const CurrentUserContextProvider = ({ children }) => {
     }
   }, [lang]);
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const value = useMemo(() => query, [query.data]);
```
dialexo, line 21: Does this really cache the query value? Isn't it passed by reference? So if `query.data` changes, the nested reference will be updated anyway, so I guess `query` is not immutable. Am I missing something?

dialexo (changes requested):
Thanks for this PR, the user menu will look slick ✨ I have a few questions on some code which I'm not sure of, and some minor style issues in tests.

src/components/common/UserSwitchWrapper.js
pyphilia: `GraaspUserSwitch` is factored out in graasp-ui, so it's not aware of the context. At first the wrapper was in compose, but since I needed the exact same component I factored it out in graasp ui as well. Maybe I could have pass the context and use `useContext` in the component? Not sure it works.

src/components/context/CurrentUserContext.js
pyphilia: I needed it because it was infinitely looping for some reasons. It's not the best solution because it won't provide the most accurate isError/isLoading... of the value. But it's anyway not supposed to change much.

src/components/common/UserSwitchWrapper.js
dialexo: Ah my bad, my explanation was a bit confusing. Obviously I don't think passing the context object itself would be a great idea, but would it make sense to consume the context first and pass the value object as a whole to the `GraaspUserSwitch` component? Something like this:

```tsx
const userQuery = useContext(CurrentUserContext)

...
return (
   <>
      <GraaspUserSwitch
        ...
        currentMemberQuery={userQuery}
        ...
      />
   <>
)
```

Not sure if it's more readable though

dialexo (approved):

PR #394, opened 2022-07-06
328/addPasswordEdition

victorggonzalez:
This PR contains changes to allow member update his/her password from the profile view


src/components/member/DeleteMemberDialog.js
```diff
@@ -20,7 +21,25 @@ const useStyles = makeStyles(() => ({
...
+    alignItems: 'flex-start',
+    margin: '12px 0px',
+  },
+  deleteAccountContainer: {
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'flex-start',
+    alignItems: 'center',
+    padding: '0px 12px',
+    margin: '0px 0px',
```
pyphilia, line 38: Same as above.

src/components/member/DeleteMemberDialog.js
```diff
@@ -64,16 +83,32 @@ const DeleteMemberDialog = ({ id }) => {
...
+            <Typography variant="h5">Delete this account</Typography>
+          </Grid>
+          <Grid
+            container
+            spacing={3}
+            className={classes.deleteAccountContainer}
+          >
+            <Typography className={classes.note}>
+              Once you delete an account, there is no going back. Please be
+              certain.
```
pyphilia, line 98: Same as above.

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+  newPasswordValidator,
+  passwordValidator,
+  strengthValidator,
+} from '../../utils/validation';
+
+const useStyles = makeStyles(() => ({
+  mainContainer: {
+    flexDirection: 'column',
+    alignItems: 'flex-start',
+    margin: '12px 0px',
```
pyphilia, line 28: Same as above.

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+  },
+  changePasswordContainer: {
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'flex-start',
+    alignItems: 'center',
+    margin: '12px 0px',
```
pyphilia, line 35: Same as above.

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+  },
+  buttonItem: {
+    display: 'flex',
+    padding: '12px',
```
pyphilia, line 39: Same as above.

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+    flexDirection: 'column',
+    justifyContent: 'flex-end',
+    alignItems: 'flex-start',
+  },
+  note: {
+    fontSize: '12px',
+    margin: '4px 0 2px',
```
pyphilia, line 46: Same as above.

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+
+  const userId = user.get('id');
+
+  const onClose = () => {
+    setCurrentPassword('');
+    setNewPassword('');
+    setConfirmPassword('');
+  };
+
+  const handleChangePassword = () => {
```
pyphilia, line 72: This function is overall difficult to understand because it lacks comments, and because you have lots of nested if clauses. I think you might take advantage of throwing errors (rather than returning strings in your validator functions), then you could try-catch the errors and set the error/notify. Returning more often would help as well (even if it's fake data, or just a boolean defining if yes or not you could change the password).

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+    setConfirmPassword(event.target.value);
+    setConfirmPasswordError(passwordValidator(event.target.value));
+  };
+
+  return (
+    <>
+      <Grid container spacing={3} className={classes.mainContainer}>
+        <Grid item xs={8}>
+          <Grid item xs={12}>
+            {user.get('password') !== null ? (
```
pyphilia, line 128: `user.get('password')` ? Your password could be empty (usually not but maybe)
And since only the text changed, maybe you do the check in `Typography`?

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+              <Typography variant="h5">{t('Set Password')}</Typography>
+            )}
+          </Grid>
+          <Grid
+            container
+            spacing={3}
+            className={classes.changePasswordContainer}
+          >
+            <Grid item xs={12} sm={6}>
+              {user.get('password') !== null ? (
```
pyphilia, line 140: Same as above.

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+                  variant="outlined"
+                  value={currentPassword}
+                  error={currentPasswordError}
+                  helperText={currentPasswordError}
+                  onChange={handleCurrentPasswordInput}
+                  id={USER_CURRENT_PASSWORD_INPUT_ID}
+                  type="password"
+                />
+              ) : (
+                <Grid item xs={12} sm={6} />
```
pyphilia, line 154: Why do you need these empty grid?
If you need to have a new line, you can nest `Grid` tags (with `container` rather than `item`)

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+              )}
+            </Grid>
+            <Grid item xs={12} sm={6} />
```
pyphilia, line 157: Same as above.

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+              <Grid item xs={12} sm={6} className={classes.buttonItem}>
+                <Typography className={classes.note}>
+                  {t('I forgot my password:')}
+                </Typography>
+                <Button
+                  id={CONFIRM_RESET_PASSWORD_BUTTON_ID}
+                  variant="outlined"
+                  disabled
+                  // TO DO:
+                  // onClick={() => handleChangePassword()}
```
pyphilia, line 211: Change the code? It was supposed to reset the password right?

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,228 @@
...
+      );
+      if (checkingPasswordUpdate) {
+        toast.error(checkingPasswordUpdate);
+      } else {
+        const checkingStrength = strengthValidator(newPassword);
+        if (checkingStrength) {
+          toast.error(checkingStrength);
+        } else {
+          onUpdatePassword({
+            id: userId,
```
pyphilia, line 100: Do you really need this id? I think you can only change your own password.

src/utils/validation.js
```diff
@@ -0,0 +1,44 @@
+import validator from 'validator';
+
+import {
+  PASSWORD_CONFIRM_ERROR,
+  PASSWORD_EMPTY_ERROR,
+  PASSWORD_EQUAL_ERROR,
+  PASSWORD_WEAK_ERROR,
+} from '../config/messages';
+
+export const strengthValidator = (password) => {
```
pyphilia, line 10: In general for this file: Since these are validators, you would expect to receive booleans, and maybe throwing errors instead of strings.

src/components/member/PasswordSetting.js
victorggonzalez: Got rid of it!

src/components/member/PasswordSetting.js
victorggonzalez: The idea is to request a password reset to the admin, which will be handled from the Admin panel. I leave it as work to do

src/components/member/PasswordSetting.js
victorggonzalez: Left only the 'Change password' text as we are not going to evaluate the password from the user

victorggonzalez: user.get('password') returns True if the user has a password set or false if not. 
I refactored this check inside `Typography ` element so that it appears based on that.

src/components/member/PasswordSetting.js
victorggonzalez: It is necessary because the task manager checks if the member is the same as the one trying to perform the update operation

src/components/member/PasswordSetting.js
victorggonzalez: I decided to return 'error strings' because that is the current approach used in graasp-auth. It is useful because TextField helperText prop shows a string whenever an error appears. Therefore, I use the error as the corresponding string. 
I did some refactoring to this function so it becomes clearer

victorggonzalez:
Thanks for comments! I resolved most of them and answered a few which we can discuss. Also included some minor changes in the AvatarSetting component to match the appearance of the container below

src/components/member/DeleteMemberDialog.js
```diff
@@ -64,16 +79,33 @@ const DeleteMemberDialog = ({ id }) => {
...
-        className={classes.deleteButton}
-        color="secondary"
-        onClick={() => setOpen(true)}
-      >
-        {t('Delete Account')}
-      </Button>
+      <Grid container spacing={3} className={classes.mainContainer}>
+        <Grid item xs={8}>
+          <Grid item xs={12}>
+            <Typography variant="h5">Delete this account</Typography>
```
pyphilia, line 85: Use `t`

pyphilia (changes requested):
Looks much better! I've answered some of the comments I've already left.

src/components/member/PasswordSetting.js
pyphilia: Indeed, but since I've commented in your other PR that you should have a dedicated endpoint for setting a password, you won't need the id anymore (since the endpoint won't depend on the user id)!

pyphilia (changes requested):
I've left some comments, regarding the password field^^

victorggonzalez:
Solved the remaining issues. I will perform the rebase once the PR is approved

src/components/member/PasswordSetting.js
```diff
@@ -0,0 +1,208 @@
...
+  const handleChangePassword = () => {
+    try {
+      // verify there are no empty inputs
+      verifyEmptyPassword();
+      // perform validation when all fields are filled in
+      newPasswordValidator(currentPassword, newPassword, confirmPassword);
+      // check password strength for new password
+      strengthValidator(newPassword);
+      // perform password update
+      onUpdatePassword({
```
pyphilia, line 93: It makes more sense if it's just called `updatePassword`

pyphilia (changes requested):
Looks good! Could you write tests? You won't test the backend behavior, but probably going to the profile, filling the fields (check if it validates correctly) and if it sends the right data.

victorggonzalez:
Just included the tests! It fills the input fields and checks the request contains the appropriate info.

PR #403, opened 2022-07-26
accessibility: add title to the pages

xiaotiansu:
Everything else seems following the rules of accessibility


pyphilia (approved):
It's all good for me, but is changing the title the only change necessary for accessibility?

src/components/RecycleBinScreen.js
```diff
@@ -60,6 +62,9 @@ const RecycleBinScreen = () => {
 
   return (
     <Main>
+      <Typography variant="h4" align="left" id={TITLE_TEXT_ID}>
+        {t('Build and share rich educational resources')}
```
pyphilia, line 66: It's the Recycle Bin screen, so perhaps we want a more adapted text?

pyphilia (changes requested):
Thanks for the PR, but not sure about adding the titles, especially if they don't fit the screens. Could you provide screenshots? It's easier for me if you change the layout.

src/components/RecycleBinScreen.js
xiaotiansu: image
image
image
image
Yeah, maybe we could have adaptive texts for each page.

xiaotiansu:
I also resolved the conflict, but it seems missing on the github
image

src/components/main/FavoriteItems.js
```diff
@@ -71,7 +73,10 @@ const FavoriteItems = () => {
 
   return (
     <Main>
-      <ItemHeader />
+      <Typography variant="h4" align="left" id={TITLE_TEXT_ID}>
+        {t('Favourite Items')}
```
pyphilia, line 77: I think we went for `Favorite`

src/config/selectors.js
```diff
@@ -1,3 +1,4 @@
+export const TITLE_TEXT_ID = 'homeTitle';
```
pyphilia, line 1: An id shouldn't be duplicated, instead create one id different for each title, and the name of the selector should reflect the string content.
```
HOME_TITLE_ID = 'homeTitle'
```

pyphilia (changes requested):
Thanks for the PR! Did you check the item page? And the settings of these items?

PR #406, opened 2022-08-02
feat: add mentions

### LINKED ISSUE #405: Add mentions and notifications

spaenleh:
This PR adds the UI for viewing and interacting with chat notifications.
For visuals refer to graasp/graasp-chatbox#46.

closes #405

### DIFF
--- a/cypress/integration/item/chatbox/chatbox.spec.js
+++ b/cypress/integration/item/chatbox/chatbox.spec.js
@@ -49,7 +49,7 @@ describe('Chatbox Scenarios', () => {
     cy.get(`#${CHATBOX_ID} #${CHATBOX_INPUT_BOX_ID} button`).click();
     cy.wait('@postItemChatMessage').then(({ request: { body } }) => {
       expect(body.body.message).to.equal(message);
-      expect(body.body.mentions).to.equal([]);
+      expect(body.body.mentions).to.deep.equal([]);
 
       // mock websocket response
       client.receive({
--- a/cypress/integration/item/chatbox/chatbox.spec.js
+++ b/cypress/integration/item/chatbox/chatbox.spec.js
@@ -48,7 +48,8 @@ describe('Chatbox Scenarios', () => {
     );
     cy.get(`#${CHATBOX_ID} #${CHATBOX_INPUT_BOX_ID} button`).click();
     cy.wait('@postItemChatMessage').then(({ request: { body } }) => {
-      expect(body.body).to.equal(message);
+      expect(body.body.message).to.equal(message);
+      expect(body.body.mentions).to.equal([]);
 
       // mock websocket response
       client.receive({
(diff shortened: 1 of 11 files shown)


src/components/context/LayoutContext.js
```diff
@@ -22,7 +22,10 @@ const LayoutContextProvider = ({ children }) => {
   const [isMainMenuOpen, setIsMainMenuOpen] = useState(true);
 
   const [isItemMetadataMenuOpen, setIsItemMetadataMenuOpen] = useState(false);
-  const [isChatboxMenuOpen, setIsChatboxMenuOpen] = useState(false);
+  // check query params to see if chat should be open
+  const chatIsOpen =
+    new URLSearchParams(window.location.search).get('chat') === 'open';
```
pyphilia, line 27: Should we have an object for the chat status? Do you plan to have many? At least let's factor it out!

pyphilia (approved):
LGTM 🍇

cypress/integration/item/chatbox/chatbox.spec.js
```diff
@@ -48,7 +48,8 @@ describe('Chatbox Scenarios', () => {
     );
     cy.get(`#${CHATBOX_ID} #${CHATBOX_INPUT_BOX_ID} button`).click();
     cy.wait('@postItemChatMessage').then(({ request: { body } }) => {
-      expect(body.body).to.equal(message);
+      expect(body.body.message).to.equal(message);
```
dialexo, line 51: Should this one also be deeply compared?

src/components/common/Chatbox.js
```diff
@@ -13,9 +13,10 @@ const { useItemChat, useMembers, useAvatar, useItemMemberships } = hooks;
 
 const Chatbox = ({ item }) => {
   const { data: chat, isLoading: isChatLoading } = useItemChat(item.id);
-  const { data: members, isLoading: isMembersLoading } = useMembers([
-    ...new Set(chat?.messages?.map(({ creator }) => creator)),
-  ]);
+  const { data: memberships } = useItemMemberships(item.id);
+  const { data: members, isLoading: isMembersLoading } = useMembers(
+    memberships?.map((m) => m.memberId)?.toArray() || [],
```
dialexo, line 18: Do all members that have a membership on the item necessarily need to be loaded for the item's chat? What if some member did not use the chat on this item?

dialexo (approved):
LGTM 🌻 I've left 2 small questions

src/components/common/Chatbox.js
spaenleh: @dialexo This is a good question, before, we did only load the members that had posted a message. But with the addition of the _mentions_, you want to be able to mention **all** members that have access to the item, for example to get their attention. If you have a better way of getting all members having access to the item, I am open to suggestions, in the mean time this is the best solution I could come up with.

cypress/integration/item/chatbox/chatbox.spec.js
spaenleh: @dialexo In this case, the `message` property is just a string, so i don't think that it needs to be deeply compared.

PR #408, opened 2022-08-04
407/ChangeImmutableImplementationToWorkProperlyWithReactQuery

alvrba:


.eslintignore
```diff
@@ -3,3 +3,4 @@ public
 coverage
 src/serviceWorker.js
 cypress/integration/examples
+src
```
pyphilia, line 6: Remove that.

src/components/common/FavoriteButton.js
```diff
@@ -35,11 +35,11 @@ const FavoriteButton = ({ item, type, onClick }) => {
 
   const handleUnfavorite = () => {
     mutation.mutate({
-      id: member.get('id'),
+      id: member.id,
       extra: {
-        favoriteItems: member
-          ?.get('extra')
-          .favoriteItems?.filter((id) => id !== item.id),
+        favoriteItems: member?.extra.favoriteItems?.filter(
```
pyphilia, line 40:
```suggestion
favoriteItems: member?.extra?.favoriteItems?.filter(
```

src/components/item/sharing/CsvInputParser.js
```diff
@@ -36,8 +36,8 @@ const useStyles = makeStyles(() => ({
 const CsvInputParser = ({ item }) => {
   const classes = useStyles();
   const { t } = useTranslation();
-  const itemId = item.get('id');
-  const itemPath = item.get('path');
+  const itemId = item.id;
+  const itemPath = item.path;
```
pyphilia, line 40: Suggestion: destructure (we can now right?)
```js
const { id:itemId, path:itemPath } = item
```

src/components/main/ItemMenu.js
```diff
@@ -76,7 +76,7 @@ const ItemMenu = ({ item, canEdit }) => {
   };
 
   const renderAuthenticatedActions = () => {
-    if (!member || member.isEmpty()) {
+    if (!member || member.toSeq().isEmpty()) {
```
pyphilia, line 79: Check member id.

src/components/main/MainMenu.js
```diff
@@ -29,7 +29,7 @@ const MainMenu = () => {
   };
 
   const renderAuthenticatedMemberMenuItems = () => {
-    if (member?.isEmpty()) {
+    if (member?.toSeq().isEmpty()) {
```
pyphilia, line 32: Check member id.

src/components/member/AvatarSetting.js
```diff
@@ -38,7 +38,7 @@ const AvatarSetting = ({ user }) => {
   const classes = useStyles();
   const { mutate: onUploadAvatar } = useMutation(MUTATION_KEYS.UPLOAD_AVATAR);
 
-  const userId = user.get('id');
+  const userId = user.id;
```
pyphilia, line 41: I see there's a conflict between undefined or not for the user, (not you fault tho) maybe let's add a `?.` here too.

src/components/table/ItemNameCellRenderer.js
```diff
@@ -21,25 +22,28 @@ const ItemNameCellRenderer = (showThumbnails) => {
...
-          <ItemIcon
-            type={item.type}
+        {showThumbnails && (
+          <Thumbnail
             id={item.id}
-            extra={item.extra}
-            name={item.name}
+            extraThumbnail={getEmbeddedLinkExtra(item.extra)?.thumbnails?.[0]}
+            maxWidth={30}
+            maxHeight={30}
```
pyphilia, line 43: No more `type`? Why add the size here?

pyphilia (changes requested):
Looks good!  🍍  A big no is ignoring the src with eslint, I think it will reveal more error then.

src/components/table/ItemNameCellRenderer.js
alvrba: Screenshot 2022-08-12 at 11 41 51
This was the Thumbnail component we were passing before the refactor (and that caused the error because it was provided as a function instead of a JSX element). I am passing the same props. I do not know what you mean with the type (we were not passing it, and there is no `type` prop in Thumbnail), and the max width and height was passed before the refactor from here.

alvrba: I am also passing now a defaultValue prop to solve the error of the function component

src/components/main/Item.js
```diff
@@ -47,29 +49,42 @@ const Item = ({ item, memberships }) => {
...
-    memberId: member?.get('id'),
+    memberId: member?.id,
   });
 
+  // We need to convert the Record<item> (immutable) to item (object)
+  // because the following components are shared between the Grid and Table views
   const Actions = (
     <>
-      {!member.isEmpty() && <FavoriteButton member={member} item={item} />}
+      {!member.toSeq().isEmpty() && (
```
pyphilia, line 82: Check member's id?

pyphilia (approved):
I've left a minor comment, but it looks good 🍓

PR #440, opened 2022-09-09
439/add export chat

### LINKED ISSUE #439: Add Export chat and download chat in item settings

spaenleh:
This PR add the "export chat" and "clear chat" button in the item settings.

A bunch of files were also moved to TS.

Types now come from `@graasp/sdk/frontend`.

A stub for `/p/item-memberships` was added. 

closes #439


package.json
```diff
@@ -12,11 +12,11 @@
...
     "@emotion/styled": "11.10.5",
-    "@graasp/chatbox": "1.0.1",
-    "@graasp/query-client": "0.1.3",
-    "@graasp/sdk": "0.3.0",
+    "@graasp/chatbox": "github:graasp/graasp-chatbox#73-update-types",
+    "@graasp/query-client": "github:graasp/graasp-query-client#202/exportChatHook",
+    "@graasp/sdk": "github:graasp/graasp-sdk#74-afine-item-type-with-discriminated-union",
     "@graasp/translations": "1.3.0",
-    "@graasp/ui": "0.9.0",
+    "@graasp/ui": "github:graasp/graasp-ui#225-update-for-exporthook-pr",
```
spaenleh, line 19: Merge these before merging this branch.

src/components/main/ItemScreen.tsx
```diff
@@ -121,7 +127,7 @@ const WrappedItemScreen = (): JSX.Element => {
     signInButtonId: ITEM_LOGIN_SIGN_IN_BUTTON_ID,
     passwordInputId: ITEM_LOGIN_SIGN_IN_PASSWORD_ID,
     modeSelectId: ITEM_LOGIN_SIGN_IN_MODE_ID,
-  })(ItemScreen as any); // todo: improve type
+  })(ItemScreen); // todo: improve type
```
pyphilia, line 130: Remove comment.

src/utils/membership.ts
```diff
@@ -66,11 +69,12 @@ export const membershipsWithoutUser = (
 ): List<ItemMembershipRecord> =>
   memberships?.filter(({ memberId }) => memberId !== userId);
 
+// todo: Is this used ? Can we remove it ?
```
pyphilia, line 72: If it is not used, you can remove.

pyphilia (changes requested):
🎆

src/components/item/ItemContent.tsx
```diff
@@ -89,7 +82,10 @@ const ItemContent: FC<Props> = ({ item, enableEditing, permission }) => {
   );
   const isEditing = enableEditing && editingItemId === itemId;
 
-  const etherpadQuery = useEtherpad(item, 'write'); // server will return read view if no write access allowed
+  const etherpadQuery: UseQueryResult<EtherpadRecord> = useEtherpad(
+    item,
+    'write',
```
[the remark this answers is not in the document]
spaenleh: This one is a bit different from the `PermissionLevel`. There is only `write` and `read` permisisons, and they are not aligned with the item permissions (player always gets `read`)

src/components/item/settings/DownloadChatButton.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+  };
+
+  return (
+    <Stack direction="row" spacing={2} alignItems="center">
+      <StyledCSVLink
+        id={DOWNLOAD_CHAT_BUTTON_ID}
+        // add a property to pass the fileName
+        // this property will have a value after clicking the button
+        data-cy-filename={filename}
```
dialexo, line 87: See comment above

src/components/item/sharing/CreateItemMembershipForm.tsx
```diff
@@ -47,12 +46,13 @@ const CreateItemMembershipForm = ({ item, members }: Props): JSX.Element => {
...
-  // todo: use an array to later allow sending multiple invitations
-  const [invitation, setInvitation] = useState<Partial<Invitation>>({
+  // use an array to later allow sending multiple invitations
+  const [invitation, setInvitation] = useState<InvitationFieldInfoType>({
+    email: '',
     permission: PermissionLevel.Read,
   });
 
-  const isInvitationInvalid = ({ email }: { email?: string }) => {
+  const isInvitationInvalid = ({ email }: { email: string }): string => {
```
dialexo, line 55: Change the function name since it does not return a boolean anymore? Also, instead of returning an empty string if the conditions are not met, what about typing the function as returning `string | undefined` or `string | null`?

src/components/item/sharing/ItemSharingTab.tsx
```diff
@@ -85,7 +88,7 @@ const ItemSharingTab = ({ item }: Props): JSX.Element => {
         todo: show only if item is pseudomized
         */}
         {/* // todo: this will change with the refactor */}
-        {getItemLoginSchema(item?.extra as any) && (
+        {getItemLoginSchema(item?.extra as { itemLogin?: ItemLogin }) && (
```
dialexo, line 91: Why is the cast still necessary here?

src/components/item/sharing/VisibilitySelect.tsx
```diff
@@ -149,7 +146,7 @@ const VisibilitySelect: FC<Props> = ({ item, edit }) => {
         deletePublishedAndPublicTags();
         postNewTag();
         // post login schema if it does not exist
-        if (!getItemLoginSchema(item?.extra)) {
+        if (!getItemLoginSchema(item?.extra as { itemLogin?: ItemLogin })) {
```
dialexo, line 149: Same as above

src/components/main/Item.tsx
```diff
@@ -61,16 +60,17 @@ const Item: FC<Props> = ({ item, memberships }) => {
     <Thumbnail
       id={item.id}
       thumbnailSrc={
-        getEmbeddedLinkExtra(extra as EmbeddedLinkItemExtra)?.thumbnails[0]
+        item.type === ItemType.LINK
+          ? getEmbeddedLinkExtra(item.extra)?.thumbnails?.first()
+          : undefined
```
dialexo, line 65: Do the apps also have thumbnails? Also, what if we want to support more types in the future? Should it be checked generically for the specific extra field?

src/utils/itemExtra.ts
```diff
@@ -93,9 +65,11 @@ export const buildDocumentExtra = (
 ): DocumentItemExtra => ({
   [ItemType.DOCUMENT]: text,
 });
-
+/**
+ * @deprecated
+ */
 export const getDocumentExtra = (
```
dialexo, line 71: Q: why is this function deprecated and not removed like the others?

tsconfig.json
```diff
@@ -23,5 +23,5 @@
     "noImplicitReturns": true
   },
   "include": ["src"],
-  "exclude": ["node_modules", "cypress", "*.test.ts", "*.test.js", "**/.*/"]
```
dialexo, line 26: Q: is there a specific reason to exclude node_modules?

dialexo (approved):
Overall LGTM, but please have a look at my comments and let me know what you think 🏖️

cypress/e2e/invitations/editInvitation.cy.js
```diff
@@ -1,17 +1,18 @@
 import { buildItemPath } from '../../../src/config/paths';
 import {
+  ITEM_MEMBERSHIP_PERMISSION_SELECT_CLASS,
   buildInvitationTableRowSelector,
   buildPermissionOptionId,
   buildShareButtonId,
-  ITEM_MEMBERSHIP_PERMISSION_SELECT_CLASS,
 } from '../../../src/config/selectors';
-import { PERMISSION_LEVELS } from '../../fixtures/enum';
+import { PERMISSION_LEVELS } from '../../fixtures/enums';
```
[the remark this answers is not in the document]
spaenleh: For this we will need to move the test files to TS. the enum was already moved to the `fixtures/enums` file and is only used in test.

cypress/e2e/item/settings/itemSettings.cy.js
```diff
@@ -64,6 +78,85 @@ describe('Item Settings', () => {
...
+
+      cy.get(`#${DOWNLOAD_CHAT_BUTTON_ID}`)
+        .should('exist')
+        .and('be.visible')
+        // download file
+        .click();
+
+      // get file name from data-cy-filename attribute and check local csv
+      cy.get(`#${DOWNLOAD_CHAT_BUTTON_ID}`)
+        .should('have.attr', 'data-cy-filename')
```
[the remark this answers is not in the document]
spaenleh: So, this is used to pass the filename to the test to verify the content. Since the file is created with the current date and time the workaround was to pass pass its name on the dom on a test attribute.

cypress/e2e/item/settings/itemSettings.cy.js
```diff
@@ -64,6 +78,85 @@ describe('Item Settings', () => {
...
+        // download file
+        .click();
+
+      // get file name from data-cy-filename attribute and check local csv
+      cy.get(`#${CLEAR_CHAT_DIALOG_ID} #${DOWNLOAD_CHAT_BUTTON_ID}`)
+        .should('have.attr', 'data-cy-filename')
+        .then((filename) => {
+          verifyDownloadedChat(
+            filename.toString(),
+            ITEM_WITH_CHATBOX_MESSAGES.chat.length,
```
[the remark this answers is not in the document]
spaenleh: if the chat is cleared, the download chat button is hidden, and thus no file should be downloadable.

cypress/e2e/item/view/viewFolder.cy.js
```diff
@@ -47,7 +47,7 @@ describe('View Folder', () => {
       i18n.changeLanguage(CURRENT_USER.extra.lang);
     });
 
-    it.only('visit Home', () => {
+    it('visit Home', () => {
```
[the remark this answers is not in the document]
spaenleh: yes that is true, but I guess it gets used in the test files to isolate a single test, which lets you debug faster. catching it in CI would be a good idea I think.

cypress/e2e/memberships/createItemMembership.cy.js
```diff
@@ -1,11 +1,11 @@
 import { buildItemPath } from '../../../src/config/paths';
 import {
-  buildShareButtonId,
   CREATE_MEMBERSHIP_FORM_ID,
   SHARE_ITEM_EMAIL_INPUT_ID,
   SHARE_ITEM_SHARE_BUTTON_ID,
+  buildShareButtonId,
 } from '../../../src/config/selectors';
-import { PERMISSION_LEVELS } from '../../fixtures/enum';
+import { PERMISSION_LEVELS } from '../../fixtures/enums';
```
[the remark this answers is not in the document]
spaenleh: Need TS in tests.

cypress/fixtures/enums.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @deprecated use graasp sdk instead
+ * this object only exists for tests
+ */
+export const MIME_TYPES = {
+  IMAGE: ['image/png', 'image/jpg', 'image/gif', 'image/jpeg'],
+  VIDEO: ['video/mp4'],
+  AUDIO: ['audio/mpeg', 'audio/mp3'],
+  PDF: ['application/pdf'],
+};
```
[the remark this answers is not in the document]
spaenleh: Yes in the end that is the goal, but for this we need to move the test files to TS. And I did not want to do that also in this PR (i think this one is already big enough).

cypress/support/index.ts
```diff
@@ -1,17 +1,3 @@
...
-  namespace Cypress {
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    interface Chainable<Subject> {
-      // todo: improve types
-      setUpApi(args: any): Chainable<Element>
-    }
-  }
-}
+
+export {};
```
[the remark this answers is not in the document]
spaenleh: I would leave it here intil we move tests to TS and check if we can delete it.

cypress/support/utils.js
```diff
@@ -30,3 +34,24 @@ export const parseStringToRegExp = (
 };
 
 export const EMAIL_FORMAT = '[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\\.[a-zA-Z0-9-.]+';
+
+const validateCsvData = (data, numMessages) => {
+  expect(data).to.have.length(numMessages);
+};
+
+const validateHeaders = (headers) => {
```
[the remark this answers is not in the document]
spaenleh: Yes good idea !

src/components/common/Chatbox.tsx
```diff
@@ -1,38 +1,47 @@
...
-  );
-  const { mutate: deleteMessage } = useMutation(
-    MUTATION_KEYS.DELETE_ITEM_CHAT_MESSAGE,
-  );
-  const { mutate: clearChat } = useMutation(MUTATION_KEYS.CLEAR_ITEM_CHAT);
+    useCurrentUserContext();
+  const { mutate: sendMessage } = useMutation<
+    unknown,
+    unknown,
+    SendMessageFunctionParamType
```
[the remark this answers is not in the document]
spaenleh: Yes, these need to be typed when using the mutation. An improvement is to direclty export a typed funciton that specifies these types alongside the keys. I think @pyphilia is working on this right now. the `unknown` s are necesary because we need to type the third type which is the `TData`. there is no way around that.

cypress/e2e/item/view/viewFolder.cy.js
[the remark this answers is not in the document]
dialexo: > yes that is true, but I guess it gets used in the test files to isolate a single test, which lets you debug faster. catching it in CI would be a good idea I think.

That's what I suggested with cypress grep 🙂

cypress/e2e/item/copy/listCopyMultiple.cy.js
```diff
@@ -118,16 +119,12 @@ describe('Copy items in List', () => {
...
+      cy.wait('@copyItems').then(({ response: { statusCode } }) => {
         // check item is still existing in parent
         itemIds.forEach((id) => {
           cy.get(`${buildItemsTableRowIdAttribute(id)}`).should('exist');
         });
-        body.forEach((item) => {
-          cy.get(`${buildItemsTableRowIdAttribute(item.id)}`).should(
-            'not.exist',
-          );
-        });
```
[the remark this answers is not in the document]
spaenleh: There was a typo in the setup code. We receive an error after the call to copyMultiple which mean there is no new items created in the response body. We were essentially checking that the items created were not present. Before the test was passing because the check did dot wait for the elements to be rendered.

src/components/common/FavoriteButton.tsx
```diff
@@ -1,33 +1,31 @@
-import { RecordOf } from 'immutable';
-
 import { IconButtonProps } from '@mui/material/IconButton';
 
-import { FC, useContext } from 'react';
+import { FC } from 'react';
 
 import { MUTATION_KEYS } from '@graasp/query-client';
-import { Item, Member } from '@graasp/sdk';
+import { ItemRecord, MemberRecord } from '@graasp/sdk/frontend';
```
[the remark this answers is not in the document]
spaenleh: The `...Record` are not subtypes of the "real" type, a Record has "more" props than their non-immutable counterparts. So we can not really (currently) do what you suggest. I also think it makes a clearer distinction between what is immutable and what is just a simple plain JS object. I know that there are opinions against that, `immer` has a different philosophy on that point, we can debate what we should do. For the moment I think using the Record types make it clearer, and we should enforce using them whenever possible.

src/components/context/CopyItemModalContext.tsx
```diff
@@ -69,4 +77,5 @@ const CopyItemModalProvider = ({
   );
 };
 
-export { CopyItemModalProvider, CopyItemModalContext };
+export const useCopyItemModalContext = (): CopyItemModalContextType =>
```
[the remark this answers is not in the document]
spaenleh: I think that with the `...Context` it is clear that this is a context hook instead of just a custom hook. No ?

src/components/item/sharing/ItemSharingTab.tsx
spaenleh: I am not sur how the `itemLogin` extra works. We may have to adjust the type to make this cast caduc, for the moment I don't think there is a way of removing it.

spaenleh: I am not sur how the `itemLogin` extra works. We may have to adjust the type to make this cast caduc, for the moment I don't think there is a way of removing it.

spaenleh: @pyphilia can you help us out here ? Is the `itemLogin` extra only available on folders ?

src/components/main/Item.tsx
spaenleh: Mhmmm no idea, I just translated the same as in the js file.

src/components/main/NewItemModal.tsx
```diff
@@ -55,9 +62,14 @@ const NewItemModal: FC<Props> = ({ open, handleClose }) => {
...
-  const [initialItem] = useState<Partial<Item<UnknownExtra>>>({});
+  const [initialItem] = useState<ItemRecord>(convertJs({}));
+
+  // todo: find a way to create this type of literal from the enum values instead of like this...
   const [updatedPropertiesPerType, setUpdatedPropertiesPerType] = useState<{
-    [key: string]: Partial<Item<UnknownExtra>>;
+    [ItemType.FOLDER]: { type: 'folder' };
+    [ItemType.APP]: { type: 'app' };
+    [ItemType.LINK]: { type: 'embeddedLink' };
+    [ItemType.DOCUMENT]: { type: 'document' };
```
pyphilia, line 72: Use `ItemType` instead instead of strings?

pyphilia (approved):

src/components/main/NewItemModal.tsx
spaenleh: This is the type for the state. It is better (I think) to specify the type with string literals so the discriminated union is understood better.

dialexo (approved):

PR #445, opened 2022-09-26
refactor: setup typescript and sentry

### LINKED ISSUE #44: Use Typescript
- enum  https://www.typescriptlang.org/docs/handbook/enums.html And we can have them in their own files.

```export const PERMISSION_LEVELS = {
  WRITE: 'write',
  READ: 'read',
  ADMIN: 'admin',
};```

### LINKED ISSUE #444: Setup Sentry source maps
It's nearly impossible to pinpoint the error from source code. It would be much better to handle source maps.

It seems much more easy to implement if we move the repo to typescript https://github.com/graasp/graasp-compose/issues/44

### PR DESCRIPTION
**Pull request by pyphilia**
This PR adds the typescript config for the project. No files are moved to ts but the goal is now to migrate files bit by bit.

close #44 
close #444

### CODE REVIEW

File: src/react-app-env.d.ts [hunk 1]
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
dialexo, line 1: What's this file for?

dialexo (approved):

spaenleh (changes requested):
Looks good ! Nice, so we can start migrating when adding new files ! 🌴
Just the types for the react packages need to be v17 !

src/react-app-env.d.ts
spaenleh: I think it is used to expose useful types from react-scripts to make it easier to work with images for example.

pyphilia: Yes, it's automatic and we should leave it, for ts.

spaenleh (approved):

PR #468, opened 2022-10-25
466 workflow update

### LINKED ISSUE #466: Update deployment sha

pyphilia:
This PR fixes some issues, and prevent type checking by react.
Instead `tsc` will type check, which should use node's `max old size`.

close #466


.husky/pre-push
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
+npx tsc --noEmit
```
spaenleh, line 4: Is there a particular reason why we would use `npx` here instead of `yarn` ?

package.json
```diff
@@ -52,19 +51,19 @@
     "validator": "13.7.0"
   },
   "scripts": {
-    "pre-commit": "yarn prettier:check && yarn lint",
+    "pre-commit": "npx tsc --noEmit && yarn prettier:check && yarn lint",
```
spaenleh, line 54: I think this should go in a `.husky/pre-commit` file with the following content:

```sh
#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"

npx tsc --noEmit && yarn prettier:check && yarn lint
```

package.json
```diff
@@ -52,19 +51,19 @@
...
+    "start": "env-cmd -f ./.env.local react-app-rewired start",
+    "start:test": "env-cmd -f ./.env.test react-app-rewired start",
+    "start:prod": "env-cmd -f ./.env.production react-app-rewired start",
+    "start:ci": "react-app-rewired -r @cypress/instrument-cra start",
+    "build": "react-app-rewired build",
+    "build:local": "env-cmd -f ./.env.local react-app-rewired build",
+    "dist": "env-cmd -f ./.env.production react-app-rewired build",
+    "dist:dev": "env-cmd -f ./.env.development react-app-rewired build",
+    "test": "env-cmd -f ./.env.test react-app-rewired test",
+    "test:once": "env-cmd -f ./.env.test react-app-rewired test --watchAll=false",
```
spaenleh, line 66: Do we use these ? I think we only use cypress tests and not react-testing library, but i may be wrong.

src/components/item/sharing/CreateItemMembershipForm.js
```diff
@@ -21,7 +21,7 @@ import {
   SHARE_ITEM_EMAIL_INPUT_ID,
   SHARE_ITEM_SHARE_BUTTON_ID,
 } from '../../../config/selectors';
-import { buildInvitation } from '../../../utils/invitation';
+import { PERMISSION_LEVELS } from '../../../enums';
```
spaenleh, line 24: Use `@graasp/sdk` ?

src/config/queryClient.js
```diff
@@ -9,7 +9,6 @@ const {
   hooks,
   useMutation,
   ReactQueryDevtools,
-  API_ROUTES,
```
spaenleh, line 12: Why the return to `js` in this case ?

spaenleh (changes requested):
🪆 Looks good, I have added some comments and suggestions

src/components/item/sharing/CreateItemMembershipForm.js
pyphilia: It's still a js file D: later...

PR #514, opened 2022-12-14
feat: add Resizable item setting

spaenleh:
This PR adds the manual resize item setting.
It should also fix #493 by filtering options to the ones containing the search string.
The prettier command was updated to also format TS/TSX files. This generated some noise that was committed under 50b6789.

Needs:
- [x] to be rebased 
- [x] https://github.com/graasp/graasp-sdk/issues/60 to be merged
- [x] tests to pass

closes #500

PR #529, opened 2023-01-12
fix: case-insensitive autocomplete comparison

### LINKED ISSUE #528: Fix case-insensitive autocomplete
The autocomplete now works on the name of the apps but it is case-sensitive.
We should make it case-insensitive.

spaenleh:
This PR fixes dumb issues created in #514:
- comparison of the strings should be made in lowercase
- resizable setting should be show when `ItemType` is `app`

closes #528

### DIFF
--- a/src/components/item/settings/ItemSettings.tsx
+++ b/src/components/item/settings/ItemSettings.tsx
@@ -183,8 +183,7 @@ const ItemSettings: FC<Props> = ({ item }) => {
         {renderPinSetting()}
         {renderChatSetting()}
         {renderCollapseSetting()}
-        {/* todo: change itemType to app */}
-        {item.type === ItemType.FOLDER && renderResizeSetting()}
+        {item.type === ItemType.APP && renderResizeSetting()}
       </FormGroup>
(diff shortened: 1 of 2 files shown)


src/components/item/settings/ItemSettings.tsx
```diff
@@ -183,8 +183,7 @@ const ItemSettings: FC<Props> = ({ item }) => {
         {renderPinSetting()}
         {renderChatSetting()}
         {renderCollapseSetting()}
-        {/* todo: change itemType to app */}
-        {item.type === ItemType.FOLDER && renderResizeSetting()}
+        {item.type === ItemType.APP && renderResizeSetting()}
```
dialexo, line 186: Quick question, do you know why this was set to folder?

dialexo (approved):
LGTM 🏋️

src/components/item/settings/ItemSettings.tsx
spaenleh: Left over by me from testing in #514 ... 😅

PR #533, opened 2023-01-20
526 intl date

### LINKED ISSUE #526: Item date is displayed in english format MM/DD/YY even when language si set to french
Capture d’écran 2023-01-11 à 10 25 31

Also the date is cropped.

pyphilia:
close #526


src/components/main/DeleteItemDialog.tsx
```diff
@@ -20,17 +20,33 @@ const descriptionId = 'alert-dialog-description';
...
+  );
 
   const onDelete = () => {
     if (itemIds.length > 1) {
       deleteItems(itemIds);
     } else {
-      deleteItem(itemIds);
+      // we still use this call because it has feedback
+      // remove when deleteItems will have feedback
+      deleteItem(itemIds as unknown as string);
```
spaenleh, line 49: Wouldn't it be better to send `itemIds[0]` instead ?

src/utils/date.ts
```diff
@@ -1,9 +1,44 @@
+/* eslint-disable import/no-duplicates */
```
spaenleh, line 1: Do we need this ?

spaenleh (changes requested):
🐷  Looks good to me, just a few comments, here and there and we should remove branch deps once the prs are merged 😃

src/components/main/DeleteItemDialog.tsx
pyphilia: Actually it's either an array or a string.. not the best I know D:

spaenleh (approved):
🧈

PR #537, opened 2023-01-25
508 default image is broken

### LINKED ISSUE #508: default image is broken
For some reason, the default thumbnail image is set to:

https://pbs.twimg.com/profile_images/1300707321262346240/IsQAyu7q_400x400.jpg

This appears to be now broken:

image

We should use the logo in the repository by default and add a regression test to ensure default thumbnails.

spaenleh:
This PR updates the default image to use the new graasp logo similar to [graasp-player/175](https://github.com/graasp/graasp-player/pull/175)

We also updated the dependencies:
- react-scripts@4 -> react-scripts@5 and all their dependencies (eslint,...)
- chatbox@1.0.1 with mention fix

closes #508


.eslintrc
```diff
@@ -26,6 +26,10 @@
     "no-restricted-syntax": "off",
     // disable the rule for all files
     "@typescript-eslint/explicit-module-boundary-types": "off",
+    "react/function-component-definition": [
+      2,
```
dialexo, line 30: what does this 2 stand for?

package.json
```diff
@@ -6,23 +6,25 @@
...
+    "@sentry/tracing": "7.33.0",
+    "@uppy/core": "3.0.4",
+    "@uppy/dashboard": "3.2.0",
+    "@uppy/drag-drop": "3.0.1",
+    "@uppy/file-input": "3.0.1",
+    "@uppy/progress-bar": "3.0.1",
+    "@uppy/react": "3.0.2",
+    "@uppy/status-bar": "3.0.1",
+    "@uppy/tus": "3.0.5",
+    "@uppy/xhr-upload": "3.0.4",
```
dialexo, line 27: Does this major update break anything in our code?

cypress/e2e/item/thumbnails/viewThumbnails.cy.js
```diff
@@ -29,10 +26,9 @@ describe('View Thumbnails', () => {
 
     // GRID
     cy.switchMode(ITEM_LAYOUT_MODES.GRID);
-    cy.wait(TABLE_ITEM_RENDER_TIME);
     cy.get(`#${buildItemCard(items[0].id)} img`)
       .should('have.attr', 'src')
-      .and('contain', DEFAULT_IMAGE_SRC);
+      .and('contain', 'data:');
```
dialexo, line 31: Should we use the exact base64 data?

dialexo (approved):
LGTM 🧇 I've left a few questions

.eslintrc
spaenleh: I think it is for "error" [see ESLint docs](https://eslint.org/docs/latest/use/configure/rules#rule-severities), but I will change it to "error" to be more specific.

package.json
spaenleh: Only change I had to make was related to [@uppy/xhr-upload `metaFields` which was renamed to `allowedMetaFields`](https://uppy.io/docs/migration-guides.html#uppy-tus-uppy-aws-s3-uppy-xhr-upload)

cypress/e2e/item/thumbnails/viewThumbnails.cy.js
spaenleh: Mhmm we "should" but this requires to add a loader to cypresse's webpack-config, if I am not mistaken, and I don't know how to do that.

spaenleh: Or we just copy paste it from the browser, which is a bit ugly no ?

dialexo: I think it could make sense. In general I think using raw data in tests in general is a good practice (you don't want logic errors that could cause bugs in the tests themselves). However I do agree that the way the image is bundled is obscure (the compression may change, and depending on the size several sizes may be generated), so it could change while out of our influence. Not sure what's best.

PR #546, opened 2023-02-06
feat: add tutorial footer menu item

### LINKED ISSUE #545: Add tutorial link

pyphilia:
close #545

Screenshot 2023-02-06 at 15 55 47

After some feedback I've been suggested to add the tutorials' link at some place in builder. I think the footer is a good place so it's not too highlighted, but I'm happy to get your feedback.

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -272,4 +272,3 @@ export const GRAASP_ASSETS_URL = process.env.REACT_APP_GRAASP_ASSETS_URL;
 // TODO: refer from specific endpoint /tutorials?
 export const TUTORIALS_LINK =
   'https://player.graasp.org/9d80d81f-ec9d-4bfb-836a-1c6b125aef2f';
-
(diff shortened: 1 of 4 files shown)


src/config/constants.ts
```diff
@@ -268,3 +268,6 @@ export const DEFAULT_LINK_SHOW_IFRAME = false;
 export const DEFAULT_LINK_SHOW_BUTTON = true;
 
 export const GRAASP_ASSETS_URL = process.env.REACT_APP_GRAASP_ASSETS_URL;
+
+export const TUTORIALS_LINK =
+  'https://player.graasp.org/9d80d81f-ec9d-4bfb-836a-1c6b125aef2f';
```
dialexo, line 273: It may make sense to have a dedicated endpoint (e.g. `/tutorials`) so that we can dymanically adjust e.g. in the DB at some point.

dialexo (approved):
LGTM 📻 maybe in the future we should make the link dynamic, so that we can change the resource or implementation transparently. It looks good otherwise IMO 👍

PR #562, opened 2023-02-18
feat: add document styles in Builder

### LINKED ISSUE #509: Add document box styles

dialexo:
Closes #509 

Implements document styles (as Alert components).

This PR also refactors `ItemContent` with better typing: the item is first checked against its typed, and each item type now has a corresponding content component. This allows only invoking hooks only when they are actually needed, and the item extras are now correctly typed according to their discriminated union type (thanks to @spaenleh's work). It also provides cleaner loading and error handling logic, since they are scoped to however the item type behaves / fails.

[Kapture 2023-02-20 at 17.01.18.webm](https://user-images.githubusercontent.com/14943421/220154536-72b063c8-d246-4460-87e6-df7e67d98408.webm)


src/components/item/ItemContent.tsx
```diff
@@ -49,201 +62,410 @@ import { useLayoutContext } from '../context/LayoutContext';
...
+}: {
+  item: EtherpadItemTypeRecord;
+}): JSX.Element => {
+  const {
+    data: etherpad,
+    isLoading,
+    isError,
+  }: UseQueryResult<EtherpadRecord> = useEtherpad(
+    item,
+    'write', // server will return read view if no write access allowed
```
pyphilia, line 335: Use `PermissionLevel`?

src/components/item/form/DocumentForm.tsx
```diff
@@ -1,16 +1,88 @@
...
+              event.target.value === ''
+                ? undefined
+                : (event.target.value as DocumentItemExtraProperties['flavor']),
+            )
+          }
+        >
+          <MenuItem value="">None</MenuItem>
+          {['info', 'warning', 'error', 'success'].map((f) => (
+            <MenuItem key={f} value={f}>
+              {f}
```
pyphilia, line 63: Should we add `i18n`? So we can also rename the flavors.

pyphilia: Also should we create an enum for the flavor values? So we can also narrow the value in the backend for example.

dialexo: > Should we add `i18n`? So we can also rename the flavors.

Yes, I will add the PR to translations.

> Also should we create an enum for the flavor values? So we can also narrow the value in the backend for example.

This is one of those type vs enum thing: currently it is defined as an union type (see related PR: https://github.com/graasp/graasp-sdk/blob/2fdaddd15d48014215192af280b47ddbcef51b3c/src/interfaces/extra.ts#L26). Should I change it into an enum? @pyphilia @spaenleh 
Some fodder for argument:
- https://www.bam.tech/article/should-you-use-enums-or-union-types-in-typescript
- https://www.becomebetterprogrammer.com/typescript-union-types-vs-enums/
- https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant

PR #573, opened 2023-02-23
556 show share menu read only mode

### LINKED ISSUE #556: Show "Sharing" menu in `read-only` mode when user is not admin
Currently non-admin users can see the "add user by email" field.
Currently non-admin users can see the dropdown to change a users permission on the item.

All these calls fail in the backend which guaranties the permission access scheme is respected.
We should:
- not show the "add member by email" field to users without the "admin" right
- only show the user permission in "read-only" mode to users without the "admin" right

MalinSvenberg:
Closes #556

The Authorized members and pending invitations tables in the sharing menu look like this now, for users with `write` access.
Screenshot 2023-02-23 at 13 55 04
Screenshot 2023-02-23 at 13 55 25

The Authenticated Members table hasn't been tested, but it should work since it works as intended for the Authorized members table.


spaenleh (changes requested):
🐰 Nice work ! Thank you for providing the screenshots !

It would be nice to add cypress tests to validate this feature 👍

pyphilia (approved):
LGTM 🐸

src/components/item/sharing/ResendInvitationRenderer.tsx
```diff
@@ -30,7 +31,12 @@ const ResendInvitationRenderer = (
     };
 
     return (
-      <Button variant="outlined" onClick={resendEmail} disabled={clicked}>
+      <Button
+        variant="outlined"
+        onClick={resendEmail}
+        disabled={clicked}
+        className={ITEM_RESEND_INVITATION_BUTTON_CLASS}
```
spaenleh, line 38: Out of curiosity is there a reason why you prefered using a `class` selector over an `id` selector ?

cypress/fixtures/invitations.js
```diff
@@ -72,3 +72,49 @@ export const ITEMS_WITH_INVITATIONS = {
...
+          id: 'ecafbd2a-5688-11eb-be93-0242ac130002',
+          itemPath: 'bcafbd2a_5688_11eb_ae93_0242ac130002',
+          permission: PERMISSION_LEVELS.WRITE,
+          memberId: MEMBERS.ANNA.id,
+        },
+        {
+          id: 'ecafbd2a-5688-11eb-be93-0242ac130004',
+          itemPath: 'bcafbd2a_5688_11eb_ae93_0242ac130002',
+          permission: PERMISSION_LEVELS.ADMIN,
+          email: MEMBERS.BOB.email,
```
spaenleh, line 99: I think memberships should give the `memberId` not the `email`.

cypress/e2e/invitations/viewInvitation.cy.js
```diff
@@ -17,16 +22,66 @@ describe('View Invitations', () => {
...
+      ).should('have.value', permission);
+
+      cy.get(
+        `${buildInvitationTableRowSelector(
+          id,
+        )} .${ITEM_RESEND_INVITATION_BUTTON_CLASS}`,
+      ).should('exist');
     });
 
     // todo: check permission
```
spaenleh, line 47: remove todo if you check the permissions ?

spaenleh (approved):
Looks good to me, I have left some comments, we can discuss !

cypress/fixtures/invitations.js
MalinSvenberg: Yes I was a bit confused about why it's the email and not memberId in the invitations, but since the other tests for invitations used this I decided to just go with it.

src/components/item/sharing/ResendInvitationRenderer.tsx
MalinSvenberg: Mostly because I didn't figure out how to send the correct id as a prop, and since I only was interested to see if there was a button or not. Also because a class selector was used for the permissions selector.

PR #575, opened 2023-02-24
fix: remove player button from item header actions

### LINKED ISSUE #574: Remove player button in item actions
Since navigation is now contextual, there is no need to have the player button in the item actions anymore as it spawns new tabs.

spaenleh:
This PR removes the player button from the item header actions.

closes #574


pyphilia (approved):
💯

PR #578, opened 2023-02-28
fix(quill): save empty string when content is empty

### LINKED ISSUE #570: Save empty content when text editor is fully cleared
## Problem:
When all the text from the quill editor is removed, the saved content of the document is not `''` but `<p><br></p>`.

So when rendered it will occupy a non-null space. 

## Solutions:
1. Strip the content before saving, to save an empty string instead of a breakline
2. Save as is but check for "empty line content" when rendering the text

spaenleh:
This PR ensures that the content saved by the text editor is an empty string when the editor is cleared.

closes #570


pyphilia (approved):
💯

PR #581, opened 2023-03-01
490 fix sentry version

### LINKED ISSUE #490: Fix Sentry version
Currently `vundefined` is sent as the app version.

spaenleh:
This PR updates the `release` field of the sentry config to send `graasp-builder@vX.Y.Z`

It uses the name of the app from the package by reading the json file.
It pulls the version from the `REACT_AP_VERSION` variable or uses `latest` as a default.

closes #490


README.md
```diff
@@ -15,6 +15,7 @@ PORT=3111
 REACT_APP_SHOW_NOTIFICATIONS=true
 REACT_APP_AUTHENTICATION_HOST=http://localhost:3001
 REACT_APP_H5P_INTEGRATION_URL=
+REACT_APP_VERSION=latest-dev
```
dialexo, line 18: Does it make sense for these to be auto-supplied? Can't we inject e.g. the latest git sha by default?
Otherwise I guess it's fine

README.md
```diff
@@ -41,6 +42,7 @@ PORT=3111
 REACT_APP_SHOW_NOTIFICATIONS=false
 REACT_APP_NODE_ENV=test
 REACT_APP_H5P_INTEGRATION_URL=
+REACT_APP_VERSION=latest-test
```
dialexo, line 45: Same as above

src/config/constants.ts
```diff
@@ -8,6 +8,7 @@ import {
 import { ITEM_LAYOUT_MODES } from '../enums';
 
 export const APP_NAME = 'Graasp';
+export const PLATFORM_NAME = 'GraaspBuilder';
```
dialexo, line 11: I found it neat that you imported the `name` field of `package.json` in `index.tsx`. Would it make sense to do the same here?

src/config/constants.ts
```diff
@@ -20,6 +21,8 @@ export const NODE_ENV =
 
 export const SENTRY_DSN = process.env.REACT_APP_SENTRY_DSN;
 
+export const APP_VERSION = process.env.REACT_APP_VERSION || 'latest';
```
dialexo, line 24: Could we inject the latest git sha also here?

src/index.tsx
```diff
@@ -8,8 +8,10 @@ import ReactGA from 'react-ga4';
 import { hasAcceptedCookies } from '@graasp/sdk';
 import '@graasp/ui/dist/bundle.css';
 
+import pkg from '../package.json';
```
dialexo, line 11: Neat 🤯

src/index.tsx
```diff
@@ -18,12 +20,13 @@ import {
 import { SENTRY_ENVIRONMENT, SENTRY_TRACE_SAMPLE_RATE } from './config/sentry';
 import './index.css';
 
+console.log(`${pkg.name}@${APP_VERSION}`);
```
dialexo, line 23: Should this `console.log` be removed?

dialexo (approved):
LGTM but please consider my suggestions 🦐

src/config/constants.ts
spaenleh: I should remove this, as it is not used now.

src/index.tsx
spaenleh: 🤙

src/config/constants.ts
spaenleh: Does it make sens to load it inside the app ? Anyway, we pass the SHA when deploying to dev in the Github actions workflow. For the default I think we can stay safe and use 'latest'.

README.md
spaenleh: We can, although it is not as easy and assumes there is a .git repo present. If you just copied the code, there is no versioning info available. And this is for local development only ... we could replace if by `dev-local`

dialexo: I see, I think it's fine then

PR #585, opened 2023-03-07
feat: add switch of actions saving

rayacers:
This PR add the switch of actions saving in the member profile and the item settings. The switch is disabled by default as we do not modify the backend yet.


pyphilia (approved):
Looks all good to me! 
I've added two minor comments, you can merge after resolving them!

PR #588, opened 2023-03-08
fix: use `CustomInitialLoader` on first load

### LINKED ISSUE #587: Use custom loader

spaenleh:
This PR uses the `CustomInitialLoader` from @graasp/ui for the first screen when the builder is fetching the member. 

It would be super cool be able to statically generate that in the `idnex.html` so there is no "white screen of death" when the bundle takes a bit longer to load.
Maybe this could be accomplished when using vite ?

closes #587


pyphilia (approved):
Nice nice 💯

PR #592, opened 2023-03-16
fix: profile picture does not show

spaenleh:
This pull-request fixes the profile picture not showing in the profile and the favorite view that has an error.


package.json
```diff
@@ -147,6 +147,7 @@
     "@mui/icons-material": "5.11.0",
     "@mui/material": "5.11.6",
     "immer": "9.0.6",
+    "@graasp/sdk": "0.9.1",
```
dialexo, line 150: I think we should remove this once sdk, query-client and ui are in sync

dialexo (approved):

PR #598, opened 2023-03-20
ci: add auto-tag workflows

spaenleh:


swouf (approved):
As for the other PRs, let's merge 💯

dialexo (changes requested):
I assume you will refactor with the extracted composite in graasp-deploy? 👍

.github/workflows/auto-tag.yml
```diff
@@ -0,0 +1,75 @@
+name: Auto Tag
```
dialexo, line 1: Do we also have a way to factor out this workflow?

dialexo (changes requested):

dialexo (approved):

PR #611, opened 2023-04-11
Improve status column display

spaenleh:


cypress/support/createUtils.js
```diff
@@ -13,7 +13,7 @@ import { ITEM_TYPES } from '../fixtures/enums';
 
 // eslint-disable-next-line import/prefer-default-export
 export const createItem = (payload, options) => {
-  cy.get(`#${CREATE_ITEM_BUTTON_ID}`).click();
+  cy.get(`#${CREATE_ITEM_BUTTON_ID}`).click({ force: true });
```
dialexo, line 16: Why is this forced? Maybe add a comment

src/components/Root.tsx
```diff
@@ -23,12 +24,13 @@ import App from './App';
 import { CurrentUserContextProvider } from './context/CurrentUserContext';
 import ModalProviders from './context/ModalProviders';
 
-const theme = createTheme(GraaspTheme);
+// const theme = createTheme(GraaspTheme);
```
dialexo, line 27: Should this be uncommented or removed?

src/components/main/ItemsTable.tsx
```diff
@@ -89,6 +94,30 @@ const ItemsTable: FC<Props> = ({
...
+    })
+    .concat(
+      itemsTags?.map(
+        (itemTags) =>
+          isItemHidden({ tags: tagList, itemTags }) ||
+          isItemPublic({ tags: tagList, itemTags }) ||
+          isItemPublished({ tags: tagList, itemTags }),
+      ),
+    )
+    .some((e) => e === true);
```
dialexo, line 119: Does this mean that the computation is done twice, i.e. once for showing the statuses, and once here to compute the inverse? In which case, would it make sense to have a partition function that traverses the computations once and splits between the rows that do have statuses, and the ones that do not?

src/components/table/BadgesCellRenderer.tsx
```diff
@@ -24,19 +26,30 @@ const BadgesCellRenderer = ({
...
         isHidden={isHidden}
+        isHiddenTooltip={t(BUILDER.STATUS_TOOLTIP_IS_HIDDEN)}
         isPublic={isPublic}
+        isPublicTooltip={t(BUILDER.STATUS_TOOLTIP_IS_PUBLIC)}
         isPublished={isPublished}
+        isPublishedTooltip={t(BUILDER.STATUS_TOOLTIP_IS_PUBLISHED)}
+        isCollapsible={isCollapsible}
+        isCollapsibleTooltip={t(BUILDER.STATUS_TOOLTIP_IS_COLLAPSIBLE)}
+        showChatbox={showChatbox}
+        showChatboxTooltip={t(BUILDER.STATUS_TOOLTIP_SHOW_CHATBOX)}
```
dialexo, line 52: IMO this shows that the current API is a bit clunky, it means that adding adding or removing badges requires changes in both the provider and the consumer.

In the future, would it make more sense to have something like:
```jsx
<ItemBadges>
   {isPinned && <Badge icon={pinIcon} tooltip={t(BUILDER.STATUS_TOOLTIP_IS_PINNED)} />}
   {isHidden && <Badge icon={hideIcon} tooltip={t(BUILDER.STATUS_TOOLTIP_IS_HIDDEN)} />}
   ... // etc
</ItemBadges>
```

dialexo (approved):
Overall LGTM, I have a few questions regarding API design as well as some nitpicks

cypress/support/createUtils.js
spaenleh: I added force, because tooltips can sometimes block the view, it looks like buttons with tooltips are kept in hover state even after clicking with cypress. Using force ensures that the click will go through even if the element is partially blocked by a tooltip.

PR #620, opened 2023-05-03
feat: adds required symbol for name of folder, document and etherpad

### LINKED ISSUE #619: Make it clear name is required when creating certain items
Right now name is required when creating a `Folder`, `Document` or `Etherpad`, this is only show through toggling the the availability of the add button. For `Etherpad` it will be clear what input is required since there is only one, but for the other two it might be missed.

MalinSvenberg:
Closes #619

Adds a * to the name text input field for `Folder`, `Document` and `Etherpad`.

Screenshot 2023-05-03 at 16 16 45

Might want use `BaseItemForm` in `EtherpadForm`?

### DIFF
--- a/src/components/item/form/BaseItemForm.tsx
+++ b/src/components/item/form/BaseItemForm.tsx
@@ -13,9 +13,15 @@ type Props = {
   updatedProperties: Partial<DiscriminatedItem<UnknownExtra>>;
   onChange: (props: Partial<DiscriminatedItem>) => void;
   item: Partial<ItemRecord>;
+  required?: boolean;
 };
 
-const BaseForm: FC<Props> = ({ onChange, item, updatedProperties }) => {
+const BaseForm: FC<Props> = ({
+  onChange,
+  item,
+  required,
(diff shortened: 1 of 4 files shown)


src/components/item/form/BaseItemForm.tsx
```diff
@@ -13,9 +13,15 @@ type Props = {
...
   onChange: (props: Partial<DiscriminatedItem>) => void;
   item: Partial<ItemRecord>;
+  required?: boolean;
 };
 
-const BaseForm: FC<Props> = ({ onChange, item, updatedProperties }) => {
+const BaseForm: FC<Props> = ({
+  onChange,
+  item,
+  required,
```
pyphilia, line 22: I don't think this is a good name. Maybe you should call this prop `isNameRequired`?

pyphilia (changes requested):
Thanks for the PR, I've left a minor comments otherwise it looks nice 🚜

dialexo (approved):

PR #627, opened 2023-05-08
fix: bump ui to 2.4.3

spaenleh:
This PR updates ui dependency to latest.

Props for the `AppItem` were updated to fit the new API.


src/components/item/ItemContent.tsx
```diff
@@ -268,17 +269,28 @@ const AppContent = ({
...
-    member={member}
     height={ITEM_DEFAULT_HEIGHT}
-    permission={permission}
-    requestApiAccessToken={Api.requestApiAccessToken}
-    context={CONTEXT_BUILDER}
+    requestApiAccessToken={(payload) =>
+      Api.requestApiAccessToken(payload, { API_HOST })
+    }
+    contextPayload={{
+      apiHost: API_HOST,
```
dialexo, line 282: Now `API_HOST` has to be passed twice, I am not sure this simplified the interface

src/components/item/ItemContent.tsx
```diff
@@ -268,17 +269,28 @@ const AppContent = ({
...
+      itemId: item.id,
```
dialexo, line 283: Not sure why `item` (line 271) and `item.id` both need to be passed as well

src/components/item/ItemContent.tsx
```diff
@@ -268,17 +269,28 @@ const AppContent = ({
...
+      memberId: member.id,
+      permission,
+      settings: item.settings,
```
dialexo, line 286: Similarly, settings has to be deref'd here

src/components/item/ItemContent.tsx
```diff
@@ -339,6 +351,7 @@ const H5PContent = ({ item }: { item: H5PItemTypeRecord }): JSX.Element => {
   return (
     <H5PItem
       itemId={item.id}
+      itemName={item.name}
```
dialexo, line 354: We should probably pass item to avoid a deref here at some point

dialexo (approved):
Ok for me, though I am not really convinced about the API change

PR #648, opened 2023-06-14
chore: update renovate.json

morganridel:
Use defaults defined here to prevent dupplication: https://github.com/graasp/renovate-config

The config can be discussed of course, once it's okay and working I will update renovate.json in all other repos.


spaenleh (approved):
🤩 Awesome !

PR #649, opened 2023-06-14
Test query client

pyphilia:
Fix some recycle tests, update query client dependency.
This will mainly fix optimistic mutations for "many items" calls that are async.


spaenleh (approved):
👌

PR #656, opened 2023-06-15
feat: refactor favorite items

morganridel:


cypress/support/server.ts
```diff
@@ -1940,3 +1942,59 @@ export const mockUpdatePassword = (
...
+};
+
+export const mockAddFavorite = (
+  _items: ItemForTest[],
+  shouldThrowError: boolean,
+): void => {
+  cy.intercept(
+    {
+      method: HttpMethod.POST,
+      url: new RegExp(`${API_HOST}/${GET_FAVORITE_ITEMS_ROUTE}/${ID_FORMAT}`),
```
pyphilia, line 1972: It doesn't have to be changes, but if possible you can use the API endpoints from query-client, so it doesn't break if we change the endpoint route.

package.json
```diff
@@ -147,7 +147,8 @@
     "nth-check": "2.1.1",
     "react-error-overlay": "6.0.11",
     "@types/react": "17.0.62",
-    "@svgr/webpack": "6.5.1"
+    "@svgr/webpack": "6.5.1",
+    "@graasp/sdk": "github:graasp/graasp-sdk#refactor-favorite"
```
pyphilia, line 151: Don't forget to change your branches ;)

src/components/common/FavoriteButton.tsx
```diff
@@ -19,48 +21,35 @@ type Props = {
...
 };
 
-export const isItemFavorite = (
+const isItemFavorite = (
   item: ItemRecord,
-  member?: MemberRecord,
-): boolean => member?.extra?.favoriteItems?.includes(item.id) || false;
-
+  favorites?: List<ItemFavoriteRecord>,
+): boolean => favorites?.map((f) => f.item.id).contains(item.id) || false;
```
pyphilia, line 27: Why not use `.some()`?

pyphilia (changes requested):
LGTM! I've left minor comments ⭐

cypress/support/server.ts
morganridel: The problem is that the endpoint builder takes a UUID to build the url, and the ID_FORMAT is a string

src/components/common/FavoriteButton.tsx
morganridel: I didn't know about it :)

PR #688, opened 2023-06-28
feat: add item menu in item header

### LINKED ISSUE #687: Add item menu in item header
It will allow to do operations from the current level (copy, move...)

pyphilia:
This PR adds the item menu button in the header. 
It's not pretty at all, but at least functional, so we don't forget it when redesigning the layout.

close #687

### DIFF
--- a/src/components/item/header/ItemHeaderActions.tsx
+++ b/src/components/item/header/ItemHeaderActions.tsx
@@ -1,6 +1,6 @@
 import { Stack } from '@mui/material';
 
-import { ItemType, PermissionLevel } from '@graasp/sdk';
+import { Item, ItemType, PermissionLevel } from '@graasp/sdk';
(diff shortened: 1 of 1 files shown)


src/components/item/header/ItemHeaderActions.tsx
```diff
@@ -70,6 +71,8 @@ const ItemHeaderActions = ({ item }: Props): JSX.Element => {
       const activeActions = (
         <>
           {showEditButton && <EditItemCaptionButton itemId={item.id} />}
+          <ItemMenu item={item.toJS() as Item} canEdit={showEditButton} />
```
spaenleh, line 74:
```suggestion
          <ItemMenu item={item.toJS() as DiscriminatedItem} canEdit={showEditButton} />
```

spaenleh (approved):
Looks good, maybe just change the type to be more accurate

PR #703, opened 2023-07-17
fix: clear new form on submit

### LINKED ISSUE #623: Clear create item when item have been added
When an item have been created and the user wants to create a new item the input from the previous one is still there. This might be harder to detect for some users (for example, users that cannot see), cause users to create unwanted duplications of the same item or be extra steps.

pyphilia:
close #623


spaenleh (approved):
LGTM ! 🍵

PR #704, opened 2023-07-17
fix: removing categories in publishing dialog

spaenleh:
Previously categories could only be deleted by de-selecting them in the drop-down.

This PR fixes the behavior to be able to use the cross icon on the tag to remove the category. Using the backspace also works to remove the latest category.

- [x] add regression test


src/components/item/publish/ItemPublishButton.tsx
```diff
@@ -79,6 +80,14 @@ const ItemPublishButton = ({
 
   return (
     <>
+      {isDisabled && (
+        <Stack direction="row" alignItems="start" gap={1} color="gray">
+          <InfoRounded fontSize="small" />
+          <Typography>
+            {/* {translateBuilder(BUILDER.LIBRARY_SETTINGS_CHILD_PUBLISHED_STATUS)} */}
```
pyphilia, line 87: Delete?

pyphilia (approved):
LGTM I've left a minor comment 🐘

PR #709, opened 2023-07-19
fix: password error shown in helpertext

spaenleh:
This PR fixes a minor bug where the password error text would be displayed inside the new password field instead of as a helper-text. This was a small typo.


pyphilia (approved):
LGTM 🐞

Issue #248, opened 2021-12-06
Allow # in url for link items


pyphilia:
allow such urls : https://example.com?ddd#here

spaenleh:
Already allowed by https://github.com/graasp/graasp-builder/blob/b4b76f639d5acd30169e7de4fd0d03785091ae64/src/utils/item.ts#L95

PR #757, opened 2023-08-21
fix(link): allow all non-whitespace chars in urls

spaenleh:
This PR proposes to simplify the validation of urls to allow some urls that currently fail. See #756 

[Test urls](https://regex101.com/r/iOZk6g/2)

@dialexo Do you think we would expose to possible security threats by allowing all non-whitespace characters in urls ?

### DIFF
--- a/src/utils/item.ts
+++ b/src/utils/item.ts
@@ -86,8 +86,9 @@ export const isUrlValid = (str?: string | null): boolean => {
       '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
       '((\\d{1,3}\\.){3}\\d{1,3})|' + // OR ip (v4) address
       'localhost)' + // OR localhost alias
-      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
-      '(\\?\\S*)?' + // query string
+      '(\\:\\d+)?' + // post (optional)
+      '(\\/\\S*?)*' + // path (lazy: takes as few as possible)
+      '(\\?\\S*?)?' + // query string (lazy)
       '(\\#\\S*)?$', // fragment locator
     'i',
   );
--- a/src/utils/item.ts
+++ b/src/utils/item.ts
@@ -87,7 +87,7 @@ export const isUrlValid = (str?: string | null): boolean => {
       '((\\d{1,3}\\.){3}\\d{1,3})|' + // OR ip (v4) address
       'localhost)' + // OR localhost alias
       '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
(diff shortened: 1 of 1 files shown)


pyphilia:
LGTM but does this comply with the back? Currently it's that:
```
.prop('url', S.string().format('url')).required(['url']),
```
Not exactly sure what it includes.

dialexo (changes requested):
> @dialexo Do you think we would expose to possible security threats by allowing all non-whitespace characters in urls ?

From the front-end POV I can at least think of social-engineered attacks similar to https://medium.com/@bobbyrsec/the-dangers-of-googles-zip-tld-5e1e675e59a5 but it's debatable whether it's an actual issue in Graasp. IMO all weaknesses can be exploited, so we should consider it a security risk.

Since the method should return a boolean, can you use the native `URL` object to perform validation? See https://snyk.io/blog/secure-javascript-url-validation/. If you want to restrict the URL to specific regexes, I think you should do both (however have a look at the regex used in the above article). Note that the `URL` method also only works against these attacks if you manually drop the undesirable bits.

In the next example, you should drop `username` and `password`, this however means that we'd then never allow URLs that contain usernames and passwords.
```
> new URL("https://github%2Ecom%E2%88%95kubernetes%E2%88%95kubernetes%E2%88%95archive%E2%88%95refs%E2%88%95tags%E2%88%95@v1271.zip/")
URL {
  href: 'https://github%2Ecom%E2%88%95kubernetes%E2%88%95kubernetes%E2%88%95archive%E2%88%95refs%E2%88%95tags%E2%88%95@v1271.zip/',
  origin: 'https://v1271.zip',
  protocol: 'https:',
  username: 'github%2Ecom%E2%88%95kubernetes%E2%88%95kubernetes%E2%88%95archive%E2%88%95refs%E2%88%95tags%E2%88%95',
  password: '',
  host: 'v1271.zip',
  hostname: 'v1271.zip',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
> 
```

Another example (not sure if this is still relevant though especially on Chrome) is using the `javascript:` URL construct to perform arbitrary XSS. One could craft an URL which executes the passed JS code, but still matches the regex.

Lastly if you replicate the change in the server as suggested by @kim you are also opening up to more possibilities in case another injection vulnerability is present further down the server request. For instance, a server which would extract the path or query string part of the URL to use it in some command (e.g. CLI or SQL) without proper validation / sanitization would allow in this regex passing arbitrary SQL or CLI operators (such as `; echo "pwned"`). So defense in depth would tell us to not accept such URLs.

So the bottom line is I think, if you only make the change on the front-end, use the `URL` constructor validation and drop what's not desirable.

pyphilia:
Should this be included in your PR? https://github.com/graasp/graasp-builder/issues/248

pyphilia:
@dialexo @spaenleh what's the state of this?

PR #763, opened 2023-08-22
Translations

pyphilia:


src/langs/constants.ts
spaenleh, line 1: General question, does it make sens to keep this file as we should have typed keys when using the t function ?

spaenleh (approved):
Looks good ! 🍇

PR #764, opened 2023-08-23
feat: add file setting maximum width

### LINKED ISSUE #646: Define a way to resize images
Images always take up the full width in builder.
Ideas for scheme:

- S, M, L with defaults (what about responsive?)
- Pixel-perfect (overflow?)
- Imported size (do users know the image resolutions?)

pyphilia:
Screenshot 2023-08-23 at 15 24 58

The max width property is available for all files. The translations will come once https://github.com/graasp/graasp-builder/pull/763 is merged. Default is Extra Large.

close #646


package.json
```diff
@@ -67,7 +67,7 @@
   },
   "scripts": {
     "dev": "vite",
-    "start": "yarn dev",
+    "start": "yarn dev --host 127.0.0.1",
```
spaenleh, line 70: Maybe we can put that setting in the vite.config.ts file and it could be changed by an env variable ? To make it more flexible.

src/components/item/settings/FileSettings.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const onChange = (e: SelectChangeEvent<MaxWidth>) => {
+    editItem({
+      id: item.id,
+      settings: { maxWidth: e.target.value as MaxWidth },
+    });
+  };
+
+  return (
+    <Box mt={4} mb={4}>
```
spaenleh, line 31: I think you can use `my` in this case.

spaenleh (approved):

Issue #787, opened 2023-09-08
Create Document Form Validation


LinaYahya:
Fix create form validation to have only name as a required field, specially that for initial form load you have to modify description to have an active add button

PR #788, opened 2023-09-08
fix create document validation

LinaYahya:
relates #787


pyphilia:
Sorry, I think we weren't on the same page, but we actually wanted to avoid saving empty documents. Maybe this PR is not necessary then? Let me know if I missed something.

LinaYahya:
@pyphilia 
The thing is I've already tried to create a document with an empty description and it always succeded, What's going right now is that we have an initial state with an inactive button, if we try to just get a dirty description input (meaning by try to add some text and delete all) then we will get an active add button!
and  I think this is wired somehow!

pyphilia (approved):

Issue #748, opened 2023-08-17
Show download button for readers


pyphilia:
Currently the download button is only available for writers/admin. 
We should let the readers see this button.
<img width="1135" alt="Screenshot 2023-09-11 at 10 42 49" src="https://github.com/graasp/graasp-builder/assets/11229627/6186558d-637b-4cec-80a6-5a615491675a">

PR #789, opened 2023-09-11
build: allow download button to reader view

LinaYahya:
Relates #748


src/components/main/ItemCard.tsx
```diff
@@ -85,19 +85,21 @@ const ItemComponent = ({
...
       )}
+      {enableEdition && (
+        <EditButton
+          item={
+            // DO NOT REMOVE cast
+            // here we cast explicitly to be equivalent to the grid which does not let us use Records
+            item.toJS() as DiscriminatedItem
+          }
+        />
+      )}
```
pyphilia, line 102: Won't this inverse their position? I think the edit button should be displayed before the download one.

pyphilia (changes requested):
Thanks for the PR, I have a question regarding the changes. Also could we have a simple test to ensure the download button is displayed for readers? As mentioned let me know if you have questions regarding the testing env 🎧

src/components/main/ItemCard.tsx
```diff
@@ -85,17 +85,19 @@ const ItemComponent = ({
...
+      {member && member.id && (
         <>
-          <EditButton
-            item={
-              // DO NOT REMOVE cast
-              // here we cast explicitly to be equivalent to the grid which does not let us use Records
-              item.toJS() as DiscriminatedItem
-            }
-          />
           <DownloadButton id={item.id} name={item.name} />
```
pyphilia, line 99: Actually, the Download button should be also available if the item is public, eg the member is not defined. You can get `isPublic` from `itemStatuses`.

cypress/e2e/item/download/downloadItem.cy.ts
```diff
@@ -0,0 +1,27 @@
+import { DOWNLOAD_BUTTON_ID } from '@/config/selectors';
+import { ITEM_LAYOUT_MODES } from '@/enums';
+
+import { SHARED_ITEMS_PATH } from '../../../../src/config/paths';
+import { SHARED_ITEMS } from '../../../fixtures/sharedItems';
+
+describe('check if download button exists for all shared items despite the creator', () => {
```
pyphilia, line 7:
```suggestion
describe('Download Item', () => {
```

and if you want add another describe scope with you current title.

cypress/e2e/item/download/downloadItem.cy.ts
```diff
@@ -0,0 +1,27 @@
...
+    cy.setUpApi(SHARED_ITEMS);
+    cy.visit(SHARED_ITEMS_PATH);
+    cy.switchMode(ITEM_LAYOUT_MODES.GRID);
+    cy.wait('@getSharedItems').then(({ response: { body } }) => {
+      for (const item of body) {
+        cy.get(`#${DOWNLOAD_BUTTON_ID}${item.id}`).should('exist');
+      }
+    });
+  });
+});
```
pyphilia, line 27: Would be nice to have a case for public then.

pyphilia (changes requested):
Sorry maybe I missed something in the previous review, I'll let you add this important change!

src/components/main/DownloadButton.tsx
```diff
@@ -37,14 +39,16 @@ export const DownloadButton = ({ id, name }: Props): JSX.Element => {
...
   return (
-    <Button
-      handleDownload={handleDownload}
-      isLoading={isDownloading}
-      title={translateBuilder(BUILDER.DOWNLOAD_ITEM_BUTTON)}
-      ariaLabel={translateBuilder(BUILDER.DOWNLOAD_ITEM_BUTTON)}
-      loaderColor="primary"
-      loaderSize={10}
-    />
+    <span id={`${DOWNLOAD_BUTTON_ID}${id}`}>
```
spaenleh, line 42: To help you could convert this to a function that takes the itemId as input and returns the string to target the element:
```ts
export const buildDownloadButtonId = (itemId: string): string => `download-button-id-${itemId}`;
```
For example. This would exist in the selectors file so you can re-use it in the tests and be sure that you have constructed your id in the same way.

cypress/e2e/item/download/downloadItem.cy.ts
spaenleh: Agree

spaenleh (changes requested):
This looks nice, I agree it would be good to have a test for the public use case. Also see my suggestion about the test id.

cypress/e2e/item/download/downloadItem.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+    cy.setUpApi(SHARED_ITEMS);
+    cy.visit(SHARED_ITEMS_PATH);
+    cy.switchMode(ITEM_LAYOUT_MODES.GRID);
+    cy.wait('@getSharedItems').then(({ response: { body } }) => {
+      for (const item of body) {
+        cy.get(`#${buildDownloadButtonId(item.id)}`).should('exist');
+      }
+    });
+  });
+  it('download button for public item should be exist', () => {
```
pyphilia, line 29:
```suggestion
  it('download button for public item should exist', () => {
```

pyphilia (approved):
LGTM I've added a minor comment but it's all good for me!

spaenleh (approved):

Issue #579, opened 2023-02-28
Show confirm dialog on delete your own membership


pyphilia:
You are free to delete your membership on an item. However there's no prevention if you missed clicked.

PR #790, opened 2023-09-11
fix: add a confirmation message when deleting membership

LinaYahya:
Relates #579 
A confirmation message was created to be shown before deleting the membership, basically, there are three different cases as follows: 
- Delete the only admin (will show a modal with a clear message without a delete button)

image

- Delete yourself (modal with a clear message with an active delete button)

image

- Delete other members (modal with a clear message with an active delete button)

image


src/components/item/sharing/ConfirmMembership.tsx
```diff
@@ -0,0 +1,92 @@
...
+    >
+      <DialogTitle id={labelId}>
+        {translateBuilder(BUILDER.DELETE_MEMBERSHIP)}
+      </DialogTitle>
+      <DialogContent>
+        <DialogContentText id={descriptionId}>{dialogText}</DialogContentText>
+      </DialogContent>
+      {isOneAdmin && memberToDelete?.permission === PermissionLevel?.Admin ? (
+        <Button onClick={handleClose} autoFocus variant="text">
+          ok
```
pyphilia, line 72: Translate this.
```suggestion
          {t('APPROVE_BUTTON_TEXT')}
```

src/components/item/sharing/ConfirmMembership.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const labelId = 'alert-dialog-title';
+const descriptionId = 'alert-dialog-description';
+
+type Props = {
+  open?: boolean;
+  handleClose: () => void;
+  item: any;
+  memberToDelete: any;
+  isOneAdmin: boolean;
```
pyphilia, line 26:
```suggestion
  hasOnlyOneAdmin: boolean;
```
Let me know if you don't like this name.

src/components/item/sharing/ConfirmMembership.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const onDelete = () => {
+    deleteItemMembership({ id: memberToDelete.instanceId, itemId: item.id });
+    handleClose();
+  };
+
+  let dialogText = '';
+  // incase of deleting the only admin
+  if (isOneAdmin && memberToDelete?.permission === PermissionLevel?.Admin) {
+    dialogText = translateBuilder(BUILDER?.DELETE_ONLY_ADMIN_ALERT_MESSAGE);
```
pyphilia, line 49:
```suggestion
    dialogText = translateBuilder(BUILDER.DELETE_ONLY_ADMIN_ALERT_MESSAGE);
```

src/components/item/sharing/ConfirmMembership.tsx
```diff
@@ -0,0 +1,92 @@
...
+  } else if (member?.id === memberToDelete?.memberInstance) {
```
pyphilia, line 50: I think it's strange to check for member existence here. Probably we want to have a check higher `if(!member)` and show an error message? Most probably you won't ever the error in real cases.

pyphilia (changes requested):
Thanks for the PR, I let you go through the comments, let me know if you have any questions 😄

src/components/item/sharing/ConfirmMembership.tsx
LinaYahya: thanks a lot @pyphilia  for all your comments, 
Regarding this point the first condition was created to check if we are going to delete the only admin, So to check first if this item has only one admin and then to check if we're going to delete this one admin!
Please correct me If I'm wrong

pyphilia: Yes! But my point is about `?.`. Probably you should have 

```
if(memberToDelete && member) {
if (isOneAdmin && memberToDelete.permission === PermissionLevel.Admin) {
    dialogText = translateBuilder(BUILDER?.DELETE_ONLY_ADMIN_ALERT_MESSAGE);
  } else if (member.id === memberToDelete.memberInstance) {
...
}
```

src/components/item/sharing/ConfirmMembership.tsx
```diff
@@ -0,0 +1,99 @@
...
+      deleteItemMembership({ id: memberToDelete?.id, itemId: item.id });
+      handleClose();
+    }
+  };
+
+  let dialogText = '';
+  // incase of deleting the only admin
+  if (
+    hasOnlyOneAdmin &&
+    memberToDelete?.permission === PermissionLevel?.Admin
```
pyphilia, line 53: See my comment (above). Don't use `?.` everywhere. Especially `PermissionLevel?.Admin` does not make sense.

src/components/item/sharing/ItemMembershipsTable.tsx
```diff
@@ -69,12 +70,24 @@ const ItemMembershipsTable = ({
 }: Props): JSX.Element => {
   const { t: translateBuilder } = useBuilderTranslation();
 
-  const { mutate: deleteItemMembership } = mutations.useDeleteItemMembership();
   const { mutate: editItemMembership } = mutations.useEditItemMembership();
   const { mutate: shareItem } = mutations.usePostItemMembership();
 
+  const [open, setOpen] = useState(false);
+  const [memberToDelete, setMemberToDelete] = useState<ItemMembership | null>(
```
pyphilia, line 77:
```suggestion
  const [membershipToDelete, setMembershipToDelete] = useState<ItemMembership | null>(
```

pyphilia (changes requested):
Thanks for the changes, I've added some more comments! Let me know if you have questions.

src/langs/en.json
```diff
@@ -248,6 +248,12 @@
   "STATUS_TOOLTIP_IS_COLLAPSIBLE": "Collapsible",
   "STATUS_TOOLTIP_SHOW_CHATBOX": "Chatbox visible",
   "SETTINGS_FILE_SETTINGS_TITLE": "File Settings",
+  "DELETE_MEMBERSHIP": "Delete Membership",
+  "DELETE_MEMBERSHIP_MODAL_CONFIRM_BUTTON": "Confirm",
+  "DELETE_OWN_MEMBERSHIP_MESSAGE": "Are you sure you want to delete your own permission on this item?",
+  "DELETE_MEMBERSHIP_MESSAGE": "Are you sure you want to delete membership on this item?",
```
spaenleh, line 254: It would be nice to include a little more information in the message. I think this information you already have in the `membershipToDelete` object. 
The values can be passed to the translation function like this: 

```ts
dialogText = translateBuilder(BUILDER.DELETE_MEMBERSHIP_MESSAGE, { name: membershipToDelete.member.name, permissionLevel: membershipToDelete.permission });
```

```suggestion
  "DELETE_MEMBERSHIP_MESSAGE": "Are you sure you want to remove {{permissionLevel}} access on this item for {{name}}? They will loose access to this item.",
```

src/langs/fr.json
```diff
@@ -247,5 +247,10 @@
   "STATUS_TOOLTIP_IS_PUBLISHED": "Publié",
   "STATUS_TOOLTIP_IS_COLLAPSIBLE": "Minifié",
   "STATUS_TOOLTIP_SHOW_CHATBOX": "Chatbox visible",
-  "SETTINGS_FILE_SETTINGS_TITLE": "Paramètres du fichier"
+  "SETTINGS_FILE_SETTINGS_TITLE": "Paramètres du fichier",
+  "DELETE_MEMBERSHIP": "Supprimer l'authorisation",
+  "DELETE_MEMBERSHIP_MODAL_CONFIRM_BUTTON": "Confirmer",
+  "DELETE_OWN_MEMBERSHIP_MESSAGE": "Êtes-vous sûr de vouloir supprimer votre propre autorisation sur cet élément ?",
```
spaenleh, line 253:
```suggestion
  "DELETE_OWN_MEMBERSHIP_MESSAGE": "Êtes-vous sûr de vouloir supprimer votre propre autorisation sur cet élément ? Vous perdrez l'accès à cet élément, cette action est irréversible.",
```

src/langs/fr.json
```diff
@@ -247,5 +247,10 @@
...
+  "DELETE_MEMBERSHIP_MESSAGE": "Êtes-vous sûr de vouloir supprimer l'autorisation à cet élément ?",
```
spaenleh, line 254:
```suggestion
  "DELETE_MEMBERSHIP_MESSAGE": "Êtes-vous sûr de vouloir supprimer l'accès en {{permissionLevel}} sur cet élément à {{name}} ? Il leur sera impossible d'accéder à cet élément.",
```

spaenleh (changes requested):
🍃 Thank you for this PR, changes look good. Please have a look at my comments.

src/components/item/sharing/ConfirmMembership.tsx
```diff
@@ -41,26 +43,52 @@ const DeleteItemDialog = ({
...
+            onClick={handleClose}
+          >
+            <CloseIcon fontSize="inherit" />
+          </IconButton>
+        }
+      >
+        <AlertTitle>Error</AlertTitle>
+        {dialogText}
+      </Alert>
+    </Snackbar>
```
spaenleh, line 90: Ah ! I did not see it like that but I guess it also works. 

Maybe if we can try to keep it consistent with the other dialogs 😅  I was thinking of putting the Alert inside the Dialog. I think it will be better to still open the dialog as this focuses the attention of the user. I feel that a Snackbar/toast is maybe a bit too small and could be out of sight on larger screens. 

What do you think ?

src/langs/de.json
```diff
@@ -230,7 +230,7 @@
   "USER_SWITCH_SWITCH_USER_TEXT": "Melden Sie sich mit einem anderen Konto an",
   "DELETE_MEMBERSHIP": "Mitgliedschaft löschen",
   "DELETE_MEMBERSHIP_MODAL_CONFIRM_BUTTON": "Bestätigen",
-  "DELETE_OWN_MEMBERSHIP_MESSAGE": "Sind Sie sicher, dass Sie Ihre eigene Berechtigung für dieses Element löschen möchten?",
-  "DELETE_MEMBERSHIP_MESSAGE": "Sind Sie sicher, dass Sie die Mitgliedschaft für diesen Artikel löschen möchten?",
-  "DELETE_ONLY_ADMIN_ALERT_MESSAGE": "Sie dürfen diesen Administrator nicht löschen, da dies der einzige Administrator ist"
+  "DELETE_OWN_MEMBERSHIP_MESSAGE": "Êtes-vous sûr de vouloir supprimer votre propre autorisation sur cet élément ? Vous perdrez l'accès à cet élément, cette action est irréversible.",
```
spaenleh, line 233: Here you put the french translation but this is the German file.

src/langs/fr.json
spaenleh: This still needs to be addressed

spaenleh (changes requested):
Almost there ! Nice proposition with the Snackbar, but I think we want to stick with the dialog.

src/components/item/sharing/ConfirmMembership.tsx
LinaYahya: I think it will be a bit strange/odd to have an alert within the dialog itself!

I feel like I'm not getting the point, would u please confirm that this is what we want to have?

image

src/langs/de.json
```diff
@@ -227,5 +227,10 @@
   "USER_SWITCH_PROFILE_BUTTON": "Siehe Profil",
   "USER_SWITCH_SIGN_OUT_BUTTON": "Abmelden",
   "USER_SWITCH_SIGNED_OUT_TOOLTIP": "Sie sind nicht angemeldet.",
-  "USER_SWITCH_SWITCH_USER_TEXT": "Melden Sie sich mit einem anderen Konto an"
+  "USER_SWITCH_SWITCH_USER_TEXT": "Melden Sie sich mit einem anderen Konto an",
+  "DELETE_MEMBERSHIP": "Mitgliedschaft löschen",
+  "DELETE_MEMBERSHIP_MODAL_CONFIRM_BUTTON": "Bestätigen",
+  "DELETE_OWN_MEMBERSHIP_MESSAGE": "Sind Sie sicher, dass Sie Ihre eigene Berechtigung für dieses Element entfernen möchten? Sie verlieren den Zugriff auf dieses Element, diese Aktion ist irreversibel.",
+  "DELETE_MEMBERSHIP_MESSAGE": "Sind Sie sicher, dass Sie {{membership}} für diesen Artikel für {{name}} entfernen möchten?",
```
spaenleh, line 234: The parameter name needs to be the same in all languages:
```suggestion
  "DELETE_MEMBERSHIP_MESSAGE": "Sind Sie sicher, dass Sie {{permissionLevel}} Zugang für diesen Artikel für {{name}} entfernen möchten?",
```

spaenleh (changes requested):
I had missed these. After this we are good.

spaenleh (approved):

PR #793, opened 2023-09-14
feat: shorten item link

### LINKED ISSUE #320: Generate short links for item's link
https://github.com/oculus42/short-uuid

LinaYahya:
shorten item link on sharing using [short-uuid](https://www.npmjs.com/package/short-uuid) for both player and builder sides. as well as to decrypt shortened uuid using params hook from graasp-ui.

closes #320


spaenleh (approved):
Thank you for this PR ! 🎉 
I was able to test locally and it all looks like it works.

Just a thought, maybe we want to use some sort of redirection when accessing a short url to expand it to the long form ? My point would be that if we have 2 different urls for the same page it could confuse users ?

PR #797, opened 2023-09-20
feat: add qr share link to share an item

### LINKED ISSUE #319: Generate QR codes for item's link

LinaYahya:
add QR code to link sharing as follows
- having an icon of QR code 
Screenshot from 2023-09-20 13-15-59

- open a modal of the generated QR Code
Screenshot from 2023-09-20 13-16-11

closes #319


src/components/common/QRCode.tsx
```diff
@@ -0,0 +1,35 @@
...
+  left: '50%',
+  transform: 'translate(-50%, -50%)',
+  width: '30%',
+  bgcolor: 'background.paper',
+  boxShadow: 24,
+  p: 4,
+};
+
+const QRCode = ({ value, open, handleClose }: Props): JSX.Element => (
+  <Modal open={open} onClose={handleClose}>
```
pyphilia, line 23: It's best to use the Dialog component https://mui.com/material-ui/react-dialog/
instead of the bare Modal.

src/components/common/QRCode.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Box sx={style}>
+      <QR
+        size={256}
+        style={{ height: 'auto', maxWidth: '100%', width: '100%' }}
```
pyphilia, line 27: What's `256`? Should it be factored out? Why do you need `width` as well?

src/components/item/sharing/SharingLink.tsx
```diff
@@ -115,46 +118,63 @@ const SharingLink = ({ itemId }: Props): JSX.Element => {
...
             </IconButton>
-          </span>
-        </Tooltip>
-      </Stack>
-    </StyledBox>
+          </Tooltip>
+        </Stack>
+      </StyledBox>
+      {link && openQrModal && (
+        <QRCode open={openQrModal} handleClose={closeQrModal} value={link} />
```
pyphilia, line 175: Would it make sense to create a `QrCodeButton` component that would contain 
```
<IconButton onClick={generateQrCode}>
              <QrCode2 />
            </IconButton>
```
and the modal?

pyphilia (changes requested):
Nice PR! I'm not quite sure about showing a bare modal with only the QR code, tho I can see the purpose of showing it big and alone. Most probably in the future we can extend the modal with a export as png button, etc.

src/components/common/QRCode.tsx
```diff
@@ -0,0 +1,46 @@
...
+          sx={{
+            position: 'absolute',
+            right: 8,
+            top: 8,
+          }}
+        >
+          <CloseIcon />
+        </IconButton>
+        <DialogContent sx={{ p: 5 }}>
+          <QR size={500} value={value} />
```
pyphilia, line 39: Why 500 now? Should this be a bit more responsive?

pyphilia (changes requested):
One last comment + it would be nice to have one test that click on this QR button and check the modal does open.

cypress/e2e/item/share/shareItem.cy.ts
```diff
@@ -126,4 +128,15 @@ describe('Share Item', () => {
       expect(url).to.include(item.id);
     });
   });
+
+  it('Share Item with QR Code', () => {
+    cy.setUpApi({ ...SAMPLE_PUBLIC_ITEMS });
+    const item = SAMPLE_PUBLIC_ITEMS.items[0] as any;
```
spaenleh, line 134: You don't need this cast here. In general it is better to not cast to any when not necessary.

```suggestion
    const item = SAMPLE_PUBLIC_ITEMS.items[0];
```

src/components/common/QRCode.tsx
```diff
@@ -0,0 +1,60 @@
...
+          }}
+        >
+          <CloseIcon />
+        </IconButton>
+        <DialogContent sx={{ p: 5 }}>
+          <Box
+            sx={{
+              width: { xs: '60vw', sm: '50vw', md: '30vw', lg: '24vw' },
+            }}
+          >
```
spaenleh, line 48: The `Box` component supports setting all sx props directly on itself:
```suggestion
          <Box width={{ xs: '60vw', sm: '50vw', md: '30vw', lg: '24vw' }}>
```

src/components/item/sharing/SharingLink.tsx
```diff
@@ -153,6 +155,11 @@ const SharingLink = ({ itemId }: Props): JSX.Element => {
             </IconButton>
           </span>
         </Tooltip>
+        {link && (
+          <span id={SHARE_ITEM_QR_BTN_ID}>
```
spaenleh, line 159: Why put the test id here and not directly on the IconButton ?

spaenleh (changes requested):
🗳️ Nice changes ! Just a few comments and it is good to go.

spaenleh (approved):

PR #802, opened 2023-09-26
feat: integrate websockets updates

dialexo:
Integrates the changes from https://github.com/graasp/graasp-query-client/pull/447


spaenleh (approved):
👌

.github/workflows/cypress.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           install: false
           # we launch the app in preview mode to avoid issues with hmr websockets from vite polluting the mocks
           start: yarn preview:test
-          browser: chrome
+          # browser: chrome
```
pyphilia, line 59: Does that change anything? Less buggy?

src/components/App.tsx
```diff
@@ -28,10 +29,14 @@ import PublishedItems from './main/PublishedItems';
 import Redirect from './main/Redirect';
 import MemberProfileScreen from './member/MemberProfileScreen';
 
+const { useItemFeedbackUpdates } = hooks;
+
 const App = (): JSX.Element => {
   const { pathname } = useLocation();
   const { data: currentMember, isLoading } = useCurrentUserContext();
 
+  useItemFeedbackUpdates?.(currentMember?.id);
```
pyphilia, line 38: Let's have a comment about why it's important to have it.

pyphilia (approved):

.github/workflows/cypress.yml
spaenleh: @pyphilia there is an open issue with cypress (actually a bug in chrome) that made the tests fail. See #806 Removing this uses the default for cypress which is Electron (i know not the best but at least it works). #807 should solve the issue with cypress.

PR #807, opened 2023-09-27
chore(deps): bump sdk, query-client and ui

spaenleh:
- fix(deps): bump sdk, ui and query-client
- fix: temporarily use electron while we fix the cypress bug


cypress.config.ts
```diff
@@ -19,12 +19,12 @@ export default defineConfig({
...
-          launchOptions.args.push('--headless=old');
-        }
-        return launchOptions;
-      });
+      // on('before:browser:launch', (browser, launchOptions) => {
+      //   if (browser.name === 'chrome' && browser.isHeadless) {
+      //     launchOptions.args.push('--headless=old');
+      //   }
+      //   return launchOptions;
+      // });
```
pyphilia, line 27: Remove?

pyphilia (approved):
LGTM 👺

cypress.config.ts
spaenleh: Yes since it worked !

PR #812, opened 2023-10-04
feat: App card's style enhancement

### LINKED ISSUE #811: App Card style Enhancement
- set an invisible outline so it does not shift when selecting the card
- make the card area extend to the bottom of the card (can be seen when hovering)
- The "Close" button can't be seen when it is on a white background ( do to un-select an app by clicking the card again).
- in the "advanced app" form it is not intuitive that we can click the "X" in the field to go back to the app list. We could have a "back" button instead.

LinaYahya:
- style enhancement includes: 
    - extend card content height
    - remove shifting down when selecting an app card
- toggle selection on card itself
- add back button from custom app selection    
    
closes #811


spaenleh:
@LinaYahya 
I took the liberty to propose some changes:
- use `outline` instead of `border` to create the selection border
  - this allows to avoid layout shifts and does not affect the box model see below the white outline when there is no border (transparent)

With border:
with_border

With outline
with_outline

- use Graasp button without re-styling for the "Back to app list"
- use a stack to create spacing between the form fields
- remove `<br />`
- fuse conditional elements controlled by `isCustomApp`
- use `Grid2` for grid layout and make number of items responsive (1 when small, 2 when medium, 3 when large)
- add a description for the custom app card
- improve `AppCard` prop to make it more generic and type the function appropriately
- added some flexbox magic to ensure the app names are at the same height (height: "100%" was not centering the card content)

spaenleh (approved):

Issue #706, opened 2023-07-17
Rework item thumbnail dispaly in item settings


spaenleh:
In Item -> Settings the thumbnail display can be improved

PR #819, opened 2023-10-11
fix(style): improve ThumbnailSettings view #706

ReidyT:
stack button and image
reduce default image size

### DIFF
--- a/src/components/item/settings/ThumbnailSetting.tsx
+++ b/src/components/item/settings/ThumbnailSetting.tsx
@@ -1,5 +1,6 @@
 import { FormEventHandler, useEffect, useRef, useState } from 'react';
 
+import { Stack } from '@mui/material';
 import Grid from '@mui/material/Grid';
 import Typography from '@mui/material/Typography';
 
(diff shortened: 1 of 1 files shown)


src/components/item/settings/ThumbnailSetting.tsx
```diff
@@ -115,14 +116,14 @@ const ThumbnailSetting = ({ item }: Props): JSX.Element | null => {
...
+  const imgUrl = thumbnailUrl ?? defaultImage;
 
   return (
     <>
       {uppy && (
         <StatusBar uppy={uppy} handleClose={handleClose} open={openStatusBar} />
       )}
-      <Grid container justifyContent="space-between">
+      <Stack spacing={3} mb={3}>
         <Grid item sm={6}>
```
spaenleh, line 127: Inside a `Stack` you should not need to use the `<Grid item>` component. You can simply have the different elements as direct children no wrapper needed.

spaenleh (changes requested):
Thank you for your first PR 🎉 

Looks good to me, but I think you can remove the useless `Grid` children.

spaenleh (approved):
🚀

PR #822, opened 2023-10-11
feat: create a dialog to downgrade own membership

LinaYahya:


src/components/item/sharing/TableRowPermissionRenderer.tsx
```diff
@@ -12,39 +30,86 @@ type TableRowPermissionRendererProps<T> = {
...
+      setOpen(false);
+    };
     const isParentMembership = useIsParentInstance({
       instance,
       item,
     });
     const onChangePermission: ItemMembershipSelectProps['onChange'] = (e) => {
       const value = e.target.value as PermissionLevel;
       // editing a parent's instance from a child should create a new instance
+      if (
```
spaenleh, line 57: Can you add a comment to explain what this condition is supposed to test ?

src/components/item/sharing/TableRowPermissionRenderer.tsx
```diff
@@ -12,39 +30,86 @@ type TableRowPermissionRendererProps<T> = {
...
+    const handleClose = () => {
+      setOpen(false);
+    };
     const isParentMembership = useIsParentInstance({
       instance,
       item,
     });
     const onChangePermission: ItemMembershipSelectProps['onChange'] = (e) => {
       const value = e.target.value as PermissionLevel;
       // editing a parent's instance from a child should create a new instance
```
spaenleh, line 56: This comment should be moved down to where it applies.

spaenleh (changes requested):
🌷 Thank you for this change.

I understand dis-allowing a user to edit their own permission is the safest way to ensure they do not loose access. But I think we should let users downgrade their permission if they want.

Just to recap:
- users that can edit memberships: only admins
- users that could downgrade their membership: only admins

Currently with the flick of a select you can remove your membership, it is too easy can have destructive and unforgiving consequences. I think we should introduce the dialog as a way to warn the user that they might do something that they can not undo. So I would change it so that they can downgrade their permission but only after having seen the dialog and accepting the consequences. 

Also important is that they should not be able to downgrade their permission if they are the last admin.

package.json
```diff
@@ -20,7 +20,7 @@
     "@graasp/query-client": "1.11.0",
     "@graasp/sdk": "1.8.0",
     "@graasp/translations": "1.19.2",
-    "@graasp/ui": "3.5.4",
+    "@graasp/ui": "https://github.com/graasp/graasp-ui.git#main",
```
spaenleh, line 23: You can use 3.6.0 which contains the update to the select component.

package.json
```diff
@@ -115,7 +115,7 @@
     "@types/node": "18.17.12",
     "@types/papaparse": "5.3.8",
     "@types/qs": "6.9.7",
-    "@types/react": "18.2.23",
+    "@types/react": "18.2.30",
```
spaenleh, line 118: this can be `18.2.31` according to `yarn why @types/react`

spaenleh (approved):
Just two comments about the version, otherwise it is good to merge once you have updated them.

PR #833, opened 2023-10-16
fix: use `??` instead of `||` for skeleton in `AppCard.tsx`

spaenleh:
This should fix the issue of the description staying in "loading" mode when it is an empty string.
With `??` it will only show the skeleton when the data is not yet available. 

Example below, the apps have loaded but the description of Task Management and Text Input are simply `""` so they show the Skeleton even if they are loaded:
Screenshot 2023-10-16 at 21 34 46


ReidyT (approved):
Everything looks good :)

LinaYahya (approved):

PR #844, opened 2023-10-23
feat: update app cards to list in new item modal

### LINKED ISSUE #835: Iterate over the design of the app list cards
# Visual Issues
- [ ] background interferes with PNG images
- [ ] icons are square, but the her banner format is too wide and crops the images

# Feedback received
- [ ] should take inspiration from the app-store list with icon text and description on a list item

pyphilia:
Screenshot 2023-10-23 at 15 54 23
Screenshot 2023-10-23 at 15 54 50

close #835


src/components/item/form/AppForm.tsx
```diff
@@ -20,21 +20,29 @@ import NameForm from './NameForm';
...
 }: AppGridProps): JSX.Element | JSX.Element[] => {
   const { useApps } = hooks;
   const { data, isLoading } = useApps();
 
   const { t: translateBuilder } = useBuilderTranslation();
 
   if (data) {
+    // filter out with search query
+    let dataToShow = searchQuery
+      ? data.filter((d) => d.name.includes(searchQuery))
```
spaenleh, line 40: We should match case insensitive

spaenleh (commented):
Nice changes ! I would like a more polished design but I see you are still iterating 😄

src/components/main/AppCard.tsx
```diff
@@ -51,33 +46,46 @@ const AppCard = ({
...
-      />
-      <CardContent sx={{ width: '100%' }}>
-        <Typography gutterBottom variant="subtitle2" component="div">
-          {name ?? <Skeleton />}
-        </Typography>
-        <Typography
-          variant="body2"
-          color="text.secondary"
+      <Stack direction="row">
+        <Stack
```
spaenleh, line 50: If you add a `flexShrink={0}` here it will fix the icon shrink issue.

src/components/main/AppCard.tsx
```diff
@@ -51,33 +46,46 @@ const AppCard = ({
...
           sx={{
-            display: '-webkit-box',
-            overflow: 'hidden',
-            WebkitBoxOrient: 'vertical',
-            WebkitLineClamp: 3,
+            height: { xs: 40, sm: 70 },
+            width: { xs: 40, sm: 70 },
```
spaenleh, line 53: You can put these directly on the Stack with the responsive syntax: `height={[40, 70]}`

src/components/main/AppCard.tsx
```diff
@@ -51,33 +46,46 @@ const AppCard = ({
...
+            background: 'white',
```
spaenleh, line 54: You can use `bgcolor="white"` directly on the stack

spaenleh (approved):
It looks good to me. 

For now using white as the background is the best, but in the end we should have all the app icons be on a color background and use white lines (like the Text Input one), so we will be able to add rounded corners etc

src/components/item/form/AppForm.tsx
```diff
@@ -147,19 +163,35 @@ const AppForm = ({ onChange, updatedProperties = {} }: Props): JSX.Element => {
...
+            placeholder={translateBuilder('Search for an app')}
+            variant="outlined"
+            size="small"
+            onChange={searchAnApp}
           />
-        </Grid2>
+          <Grid2
+            container
+            spacing={2}
+            maxHeight={400}
```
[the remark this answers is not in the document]
pyphilia: The modal will have a maximum height and the scroll will adapt, but we see a big height difference between "app" and "link" for example when you switch, I don't really like it, especially because if moves a little bit the menu on the right. Let me know what you think.

src/components/main/AppCard.tsx
pyphilia: Oh! I didn't know. In the end using a grid was working much better with typography wrapping..

Issue #895, opened 2023-12-05
Short link Library is displayed twice the first time


ReidyT:
When an item is published and you go to the sharing parameters, you will see the short link's library twice.

![image](https://github.com/graasp/graasp-builder/assets/147397675/b6cc1959-0ca6-4984-a2ab-6df778be21b6)

But once you refresh the page, it will display only once.

<img width="1441" alt="image" src="https://github.com/graasp/graasp-builder/assets/147397675/61da2df4-bb7a-429a-9d94-73a1b13b20b6">

PR #896, opened 2023-12-05
fix: reset the item platforms to avoid multiple library links

ReidyT:
See the issue #895 for more information.

### DIFF
--- a/src/components/item/sharing/shortLink/ShortLinksRenderer.tsx
+++ b/src/components/item/sharing/shortLink/ShortLinksRenderer.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { Box, Dialog, Stack } from '@mui/material';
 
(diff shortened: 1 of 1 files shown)


src/components/item/sharing/shortLink/ShortLinksRenderer.tsx
```diff
@@ -44,73 +43,51 @@ const ShortLinksRenderer = ({
...
   const [modalOpen, setOpen] = useState(false);
   const [isNew, setIsNew] = useState(false);
   const [initialAlias, setInitAlias] = useState<string>('');
   const [initialPlatform, setInitPlatform] = useState<ShortLinkPlatform>(
     Context.Player,
   );
-  // Lists the available short links platforms for the current item.
-  const [itemPlatforms, setItemPlatforms] = useState<ShortLinkPlatformType[]>([
+
+  // List of available platforms, the order is matter
```
spaenleh, line 53:
```suggestion
  // List of available platforms, the order matters
```

src/components/item/sharing/shortLink/ShortLinksRenderer.tsx
```diff
@@ -44,73 +43,51 @@ const ShortLinksRenderer = ({
...
+  const shortLinks: ShortLinkType[] = platforms
+    .map<ShortLinkType | undefined>((platform) => {
+      if (!publishedEntry && platform === ShortLinkPlatformConst.library) {
+        return undefined;
+      }
+      const clientHostManager = ClientHostManager.getInstance();
+      const url = clientHostManager.getItemAsURL(platform, itemId);
+
+      const shortLink = {
+        linkId: randomString(SHORT_LINK_ID_MAX_LENGTH),
```
spaenleh, line 69: Is this only used as a key for react ? You could use the platform for the key I think. Because here it will change every time which is contrary to the purpose of a key in react ...

spaenleh (changes requested):
Nice, just a suggestion to use the platform as the key.

PR #910, opened 2023-12-06
build: move menu items tree

LinaYahya:
Build move menu items tree to not using mui/ tree view, Basically the user will have a home as a root of the menu which contains own and shared items, 
For every folder that has children within there will be an arrow to let the user navigate to this item.  

image


src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+  return (
+    <Dialog
+      onClose={handleClose}
+      aria-labelledby={dialogId}
+      open={open}
+      scroll="paper"
+    >
+      <DialogTitle id={dialogId}>{title}</DialogTitle>
+      <DialogContent sx={{ height: '250px' }}>
```
pyphilia, line 208: Let's not fix the height. I notice you have the flickering with the width as well anyway.

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+        <Stack spacing={2} mb={2}>
+          <Breadcrumbs separator="›" aria-label="breadcrumb">
+            {paths.map((ele, index) => (
```
pyphilia, line 211: Use `parents` here.

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+  defaultParent?: DiscriminatedItem;
+}
+
+const OwnedItemsTree = ({
+  setPaths,
+  setSelectedId,
+  selectedId,
+  defaultSelectedSubItem,
+  itemIds,
+  defaultParent,
```
pyphilia, line 52: Why can't this be called `parentItem`?

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+  // serious of breadcrumbs
+  const [paths, setPaths] = useState<DiscriminatedItem[]>([]);
+
+  const { data: parentItem } = hooks.useItem(
+    selectedId === TREE_MODAL_MY_ITEMS_ID ? '' : selectedId,
+  );
+  const { data: parents } = hooks.useParents({
+    id: itemIds?.[0],
+  });
+  const [defaultSelectedSubItem, setDefaultSelectedSubItem] =
```
pyphilia, line 187: In this component, this doesn't need to be called `default`right?

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+        <CancelButton onClick={handleClose} />
+        <Button
+          onClick={onClickConfirm}
+          disabled={!parentItem && selectedId !== TREE_MODAL_MY_ITEMS_ID}
+          id={TREE_MODAL_CONFIRM_BUTTON_ID}
+          variant="contained"
+        >
+          {translateBuilder(BUILDER.MOVE_BUTTON)}
+          {(selectedId === TREE_MODAL_MY_ITEMS_ID &&
+            ` to ${translateBuilder(BUILDER.HOME_TITLE)}`) ||
```
pyphilia, line 251: Let's translate `to` as well.

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+            (parentItem?.name && ` to ${parentItem?.name}`)}
```
pyphilia, line 252: Same comment as above.

src/components/main/RowMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+        ?.split('.')
+        .map((ele) => ele.replaceAll('_', '-'));
+
+      const isItemChild = pathsSerious.indexOf(itemIds[i]);
+      if (isItemChild > -1) {
+        return isItemChild > -1;
+      }
+    }
+  }
+  return false;
```
pyphilia, line 44: You're disabling the option if the folder is a parent right? 
```suggestion
const itemPaths = itemIds.map((ele) => ele.replaceAll('_', '-'));
return itemPaths.some(item => item.path.includes(itemPath))
```
Be careful not to use `?.` everywhere, here you assume `itemPath` exist by typing it.

src/components/main/RowMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+    return null;
+  }
+  return (
+    <Button
+      onMouseEnter={handleHover}
+      onMouseLeave={handleUnhover}
+      onClick={() => {
+        setSelectedId(ele?.id);
+      }}
+      id={`${TREE_MODAL_MY_ITEMS_ID}-${ele.id}`}
```
pyphilia, line 74: Factor out in `selectors`.

src/components/main/RowMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+      sx={{
+        display: 'flex',
+        color: 'black',
+        width: '100%',
+        justifyContent: 'space-between',
+        marginBottom: '12px',
```
pyphilia, line 80: Can you use the `theme`?

src/components/main/RowMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+        background: selectedId === ele.id ? 'rgba(80, 80, 210, 0.08)' : 'none',
```
pyphilia, line 81: Factor out the color.

src/components/main/RowMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <Box display="flex" gap="4px">
+        <FolderIcon />
+        {ele.name}
+      </Box>
+      {(isHoverActive || selectedId === ele.id) && (
+        <Box display="flex">
+          <Button sx={{ padding: '0' }}>
+            {translateBuilder(BUILDER.MOVE_BUTTON)}
+          </Button>
+          {true && (
```
pyphilia, line 97: Remove??

src/components/main/RowMenu.tsx
```diff
@@ -0,0 +1,115 @@
...
+            <IconButton
+              sx={{ padding: '0' }}
+              onClick={(e) => {
+                e.stopPropagation();
```
pyphilia, line 101: Remove?

src/langs/en.json
```diff
@@ -290,5 +290,6 @@
   "SHORT_LINK_MIN_CHARS_ERROR": "The short link must at least have {{data}} chars.",
   "SHORT_LINK_MAX_CHARS_ERROR": "The short link must not exceed {{data}} chars.",
   "SHORT_LINK_INVALID_CHARS_ERROR": "The short link contains invalid chars ({{data}}).",
-  "SHORT_LINK_UNKNOWN_ERROR": "An unknown error occurred."
+  "SHORT_LINK_UNKNOWN_ERROR": "An unknown error occurred.",
+  "EMPTY_FOLDER_CHILDREN_FOR_THIS_ITEM": "No Sub-Folders For this Item"
```
pyphilia, line 294:
```suggestion
  "EMPTY_FOLDER_CHILDREN_FOR_THIS_ITEM": "This item does not contain any folder."
```

pyphilia (changes requested):
The modal tree is quite complex, I'd like more explanation and I'll review again. Thanks! 🎄

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+  itemIds = [],
+}: TreeModalProps): JSX.Element => {
+  const { t: translateBuilder } = useBuilderTranslation();
+
+  const [selectedId, setSelectedId] = useState<string>('');
+  // serious of breadcrumbs
+  const [paths, setPaths] = useState<DiscriminatedItem[]>([]);
+
+  const { data: parentItem } = hooks.useItem(
+    selectedId === TREE_MODAL_MY_ITEMS_ID ? '' : selectedId,
```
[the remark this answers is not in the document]
LinaYahya: `useItem` only receive string or undefined

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,259 @@
...
+  onClose,
+  onConfirm,
+  open = false,
+  itemIds = [],
+}: TreeModalProps): JSX.Element => {
+  const { t: translateBuilder } = useBuilderTranslation();
+
+  const [selectedId, setSelectedId] = useState<string>('');
+  // serious of breadcrumbs
+  const [paths, setPaths] = useState<DiscriminatedItem[]>([]);
```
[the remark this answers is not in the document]
LinaYahya: I don't prefer to useParents as I have a custom items such as `root` and `home`, and there's a case's that might not connected to the first parent, as an example when we want to move item from nested parent

src/components/common/MoveButton.tsx
```diff
@@ -25,36 +30,78 @@ type MoveButtonProps = {
...
+    // change item's root id to null
+    const newPayload = {
+      ...payload,
+      to:
+        payload.to &&
+        payload.to !== TREE_MODAL_MY_ITEMS_ID &&
+        validate(payload.to)
+          ? payload.to
+          : undefined,
+    };
```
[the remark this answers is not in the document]
LinaYahya: this case will not happen as there's always a value for place to move, But this is only as a kind of validation

pyphilia (changes requested):

src/components/main/MoveTreeModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+                  setPaths((prevPaths) => {
+                    const trimmedIndex = prevPaths.indexOf(ele);
+                    if (trimmedIndex === 0) {
+                      return prevPaths.slice(0, -prevPaths.length + 1);
+                    }
+                    return prevPaths.slice(0, trimmedIndex + 1);
+                  });
+
+                  if (ele.id === ROOT_MODAL_ID) {
+                    setSelectedId('');
```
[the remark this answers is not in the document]
LinaYahya: Actually the initial value is empty string So I'd prefer to keep it as it's default

PR #914, opened 2023-12-10
Add Teamwork hierarchically structure service in graasp builder

marcelmmc:


src/components/item/sharing/CsvInputParser.tsx
```diff
@@ -43,46 +46,46 @@ const CsvInputParser = ({ item }: Props): JSX.Element => {
...
-                ...(d as Partial<Invitation>),
-                itemPath,
-              }),
-            ) as any;
-
-            share({ data: dataWithItemPath, itemId });
+        const formData = new FormData();
+        formData.append("file", file);
+        // TO-DO: this request needs to be moved to query-client
+        axios.post(`${API_HOST}/items/${itemId}/invitations/upload_csv?id=${itemId}&template_id=${idTemplate}`, formData, {
```
swouf, line 83: You need to do something with the answer to this request. You should provide the user with a feedback of the operation.

src/components/item/sharing/CsvInputParser.tsx
```diff
@@ -43,46 +46,46 @@ const CsvInputParser = ({ item }: Props): JSX.Element => {
...
+  const { openModal: openMoveModal, selId: idTemplate, cleanItemSel: cleanId } = useContext(SelectItemModalContext);
+
   const handleClose = () => {
+    if (cleanId)
+      cleanId();
     setIsOpen(false);
   };
 
+  const { data: itemObj } = useItem(idTemplate)
   const handleFileChange = (e: ChangeEvent<HTMLInputElement>) => {
```
swouf, line 75: I think you should store the file in the browser but only send it when the user confirm the action.

src/components/item/sharing/CsvInputParser.tsx
```diff
@@ -43,46 +46,46 @@ const CsvInputParser = ({ item }: Props): JSX.Element => {
   const { t: translateBuilder } = useBuilderTranslation();
   const { t: translateMessages } = useMessagesTranslation();
   const { t: translateCommon } = useCommonTranslation();
-  const { id: itemId, path: itemPath } = item;
+  // const { id: itemId, path: itemPath } = item;
+  const { id: itemId } = item;
   const [isOpen, setIsOpen] = useState(false);
   const {
-    mutate: share,
+    // mutate: share,
```
swouf, line 53: Remove if it's not required.

swouf (changes requested):
Thank you for this PR.

The main issue in the interaction is that the selectec CSV is automatically uploaded, without any feedback. Once a file has been selected, you should indicate it in the UI (put the name of the file in a field under the button) and the user must confirm (_Invite_ or _Confirm_ button at the bottom of the dialog). Currently, if one choose a CSV, it can't choose a template afterwards, even if the button is positionned below. This situation doesn't make sense and naturally brings the user to do many errors.

Also, the dialog should close automatically once the action is complete.

Finally, you should provide more feedback to the user. You can either display an alert in the dialog if something fails or use a toast to display any error or success after the dialog has been closed.

src/components/item/sharing/CsvInputParser.tsx
marcelmmc: Sorry for the confusion, I am querying the name of the template folder here.

PR #931, opened 2024-01-05
feat: disable options for downgrading the only admin

### LINKED ISSUE #902: Forbidden permissions should be grayed out
In the "Share" dialog a user can't downgrade their permission (already has a special dialog for this). But we should make it clear on the permission dropdown that the "read" and "write" permissions can't be selected by disabling them.

LinaYahya:
closes #902

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "@graasp/query-client": "2.2.1",
     "@graasp/sdk": "3.3.0",
     "@graasp/translations": "1.21.1",
-    "@graasp/ui": "4.2.0",
+    "@graasp/ui": "https://github.com/graasp/graasp-ui.git#630-select-disable",
     "@mui/icons-material": "5.14.19",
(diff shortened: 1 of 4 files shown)


spaenleh (changes requested):
Thank you for this PR 🧩 

Please see my comments. I had a bit of trouble understanding how the disabled permissions are computed. Could we have maybe more comments and examples of input outputs that are expected, i.e. if the user has permission write the admin value should be disabled, if the user is the last admin all permissions should be disabled.

spaenleh:
@LinaYahya I propose some changes. 
I think that disabling all the options but allowing to click on the select makes no sens, so I added an option to completely disable the select when editing the only admin.

I also found a bug with the useMemo dependencies that made the select not update its state when adding/removing another admin. This has been fixed.

spaenleh (approved):

PR #936, opened 2024-01-10
feat: filter memberships for same member to render highest permission only

### LINKED ISSUE #729: Duplicate membership in child item
If you have a membership on the parent and the child, both will be display in the membership table. Probably we only want to display the closest one.

LinaYahya:
closes #729

### DIFF
--- a/src/components/item/sharing/ItemSharingTab.tsx
+++ b/src/components/item/sharing/ItemSharingTab.tsx
@@ -6,6 +6,7 @@ import Typography from '@mui/material/Typography';
 import {
   DiscriminatedItem,
   ItemMembership,
+  PermissionLevelCompare,
   isPseudonymizedMember,
 } from '@graasp/sdk';
 import { Loader } from '@graasp/ui';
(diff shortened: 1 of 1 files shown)


src/components/item/sharing/ItemSharingTab.tsx
```diff
@@ -33,6 +34,28 @@ type Props = {
...
+  for (const membership of memberships) {
+    const { member, permission } = membership;
+
+    if (
+      !permissionMap[member.id] ||
+      PermissionLevelCompare.gt(permission, permissionMap[member.id].permission)
+    ) {
+      permissionMap[member.id] = membership;
+    }
+  }
```
pyphilia, line 55: You can optimize with a `.reduce` function. I guess the accumulator could look like `{...acc, [member.id]: membership }`

pyphilia (approved):

src/components/item/sharing/ItemSharingTab.tsx
```diff
@@ -33,6 +34,27 @@ type Props = {
...
 };
 
+interface PermissionMap {
+  [key: string]: ItemMembership;
+}
+
+const reduceMembershipsForSameMember = (
+  memberships: ItemMembership[],
+): ItemMembership[] => {
+  const permissionMap = memberships.reduce((acc: PermissionMap, curr) => {
```
spaenleh, line 44: FYI, this form also works, both are equivalent:
```suggestion
  const permissionMap = memberships.reduce<PermissionMap>((acc, curr) => {
```

src/components/item/sharing/ItemSharingTab.tsx
```diff
@@ -33,6 +34,27 @@ type Props = {
   memberships?: ItemMembership[];
 };
 
+interface PermissionMap {
+  [key: string]: ItemMembership;
+}
+
+const reduceMembershipsForSameMember = (
```
spaenleh, line 41: Nitpicking: I would change the name of this function to something like: `selectHighestMemberships`. The `ForSameMember` feels a bit weird. It should read naturally when you use it in a component prop.

spaenleh (approved):
Looks good. I just provide some comments.

spaenleh (approved):

pyphilia (approved):

PR #952, opened 2024-01-15
fix: disable permission props

LinaYahya:


src/components/item/sharing/ItemMembershipSelect.tsx
```diff
@@ -71,9 +56,9 @@ const ItemMembershipSelect = ({
       values={Object.values(PermissionLevel).map((v) => ({
         value: v,
         text: enumT(v),
-        disabled: disabledMap[v],
+        disabled,
       }))}
-      disabled={typeof disabled === 'boolean' ? disabled : undefined}
+      disabled={disabled}
```
spaenleh, line 61: This is ignored by the current version of UI. See https://github.com/graasp/graasp-ui/issues/646

src/components/item/sharing/ItemMembershipSelect.tsx
```diff
@@ -71,9 +56,9 @@ const ItemMembershipSelect = ({
       values={Object.values(PermissionLevel).map((v) => ({
         value: v,
         text: enumT(v),
-        disabled: disabledMap[v],
+        disabled,
```
spaenleh, line 59: I don't like that we have to disable all options for the select to be disabled. This is not clean.

spaenleh (changes requested):
Thank you for noticing the bug. 🪲 

I would have liked to see a description of the bug/problem in an issue first.
Also there is no tests in this PR that guaranties the behaviour os tested for future changes. 
When proposing changes, it is better for the readability of the PR to start fresh from an up-to-date main branch. 

This feels more like a band-aid fix. As it does not handle the root issue.
If you remove line 59, it does not work as expected.

PR #960, opened 2024-01-18
fix: only show admin and write accessible in move modal

pyphilia:
Merge this first: https://github.com/graasp/graasp-query-client/pull/577


spaenleh (approved):
Nice !

PR #971, opened 2024-01-24
feat(ui): create separate pages for item share, publish and settings

### LINKED ISSUE #955: Create page for share, publish and settings

LinaYahya:
image

- create separate pages for sharing, publishing, settings, and information for an item.
- create an information page and move the update thumbnail to info page instead of settings page 
closes #955


src/components/App.tsx
```diff
@@ -82,7 +91,13 @@ const App = (): JSX.Element => {
           path={PUBLISHED_ITEMS_PATH}
           element={<PublishedWithAuthorization />}
         />
-        <Route path={buildItemPath()} element={<ItemScreen />} />
+        <Route path={buildItemPath()} element={<ItemScreenLayout />}>
+          <Route index element={<ItemScreen />} />
+          <Route path={SHARE_PATH} element={<ShareItemScreen />} />
+          <Route path={PUBLISH_PATH} element={<PublishItemScreen />} />
+          <Route path={SETTINGS_PATH} element={<SettingsItemScreen />} />
+          <Route path={INFORMATION_PATH} element={<ItemInformationScreen />} />
```
pyphilia, line 99: Let's add `ITEM_` to all path variable names. Eg `ITEM_SHARE_PATH`

src/components/common/ItemMetadataButton.tsx
```diff
@@ -1,37 +1,27 @@
...
-import {
-  ITEM_INFORMATION_BUTTON_ID,
-  ITEM_INFORMATION_ICON_IS_OPEN_CLASS,
-} from '../../config/selectors';
+import { ITEM_INFORMATION_BUTTON_ID } from '../../config/selectors';
 import { BUILDER } from '../../langs/constants';
-import { useLayoutContext } from '../context/LayoutContext';
 
-const ItemMetadataButton = (): JSX.Element => {
+const ItemMetadataButton = ({ id }: { id: string }): JSX.Element => {
```
pyphilia, line 12: Maybe let's name it `itemId` to avoid confusion.

src/components/item/settings/ItemSettings.tsx
```diff
@@ -203,6 +211,12 @@ const ItemSettings = ({ item }: Props): JSX.Element => {
 
   return (
     <Container disableGutters sx={{ mt: 2 }}>
+      <Alert severity="warning" sx={{ marginBottom: 2 }}>
+        {translateBuilder(BUILDER.UPDATE_THUMBNAIL_AT_INFO_ALERT)}{' '}
+        <Link to={buildItemInformationPath(item.id)}>
+          {translateBuilder(BUILDER.UPDATE_THUMBNAIL_LINK_TEXT)}
```
pyphilia, line 217: I think it's possible to use only one component here using this https://react.i18next.com/latest/trans-component

src/components/item/settings/ThumbnailSetting.tsx
```diff
@@ -149,6 +137,18 @@ const ThumbnailSetting = ({ item }: Props): JSX.Element | null => {
           maxWidth={THUMBNAIL_SETTING_MAX_WIDTH}
           maxHeight={THUMBNAIL_SETTING_MAX_HEIGHT}
         />
+        <input
+          type="file"
+          accept="image/*"
+          onInput={onSelectFile}
+          // onChange is successfully triggered in test
```
pyphilia, line 144: What about this comment? not sure what information it is. If it was a previous comment I guess you can remove it.

src/components/main/ItemScreens/ItemScreen.tsx
```diff
@@ -0,0 +1,25 @@
...
+import ItemMain from '@/components/item/ItemMain';
+
+import { OutletType } from './type';
+
+const ItemPage = (): JSX.Element => {
+  const { item, permission } = useOutletContext<OutletType>();
+
+  return (
+    <>
+      {item.type === ItemType.FOLDER ? <FileUploader /> : undefined}
```
pyphilia, line 17:
```suggestion
      {item.type === ItemType.FOLDER && <FileUploader />}
```

src/components/main/ItemScreens/ItemScreen.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { ItemType } from '@graasp/sdk';
+
+import FileUploader from '@/components/file/FileUploader';
+import ItemContent from '@/components/item/ItemContent';
+import ItemMain from '@/components/item/ItemMain';
+
+import { OutletType } from './type';
+
+const ItemPage = (): JSX.Element => {
+  const { item, permission } = useOutletContext<OutletType>();
```
pyphilia, line 13: Maybe `ItemContent` can use `useOutletContext` itself, so you don't need to pass all arguments from `ItemScreen`.

src/components/main/ItemScreens/ItemScreenLayout.tsx
```diff
@@ -0,0 +1,65 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [itemId]);
+
+  if (isLoading || isLoadingItemMemberships) {
+    return <Loader />;
+  }
+
+  if (!itemId || !item || isError || isErrorItemMemberships) {
+    return <ErrorAlert />;
+  }
```
pyphilia, line 43: Basile told me it's better to have `isLoading` and errors at the end of the component. So instead you would have
```
if(item) {
  const itemMembership = ...
}

if(isLoading) { ... }

return <ErrorAlert />;
```

src/components/main/ItemScreens/PublishItemScreen.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <Box py={1} px={2}>
+      <Navigation />
+      <Box px={2}>
+        <Button
+          startIcon={<ArrowBack fontSize="small" />}
+          variant="text"
+          onClick={() => navigate(buildItemPath(itemId))}
+        >
+          {translateBuilder(BUILDER.BACK)}
+        </Button>
```
pyphilia, line 35: I think you can create a component for the back button and reuse it.

src/components/main/ItemScreens/SettingsItemScreen.tsx
```diff
@@ -0,0 +1,52 @@
...
+            onClick={() => navigate(buildItemPath(itemId))}
+          >
+            {translateBuilder(BUILDER.BACK)}
+          </Button>
+          <ItemSettings item={item} />
+        </Box>
+      </Box>
+    );
+  }
+  return <Navigate to={buildItemPath(itemId)} replace />;
```
pyphilia, line 46: Maybe add a comment that this redirects the user to the item if he doesn't have the permission to access this page. Same for all other pages.

src/langs/en.json
```diff
@@ -302,5 +302,8 @@
   "TO_FOLDER": "to {{name}}",
   "ROOT": "Root",
   "ITEM_SELECTION_NAVIGATION_RECENT_ITEMS": "Recent Items",
-  "ITEM_SELECTION_NAVIGATION_PARENT": "Move above"
+  "ITEM_SELECTION_NAVIGATION_PARENT": "Move above",
+  "BACK": "Back",
+  "UPDATE_THUMBNAIL_AT_INFO_ALERT": "Update thumbnail moved to information page,",
```
pyphilia, line 307:
```suggestion
  "UPDATE_THUMBNAIL_AT_INFO_ALERT": "You can change the item's thumbnail in its information page,",
```

src/langs/en.json
```diff
@@ -302,5 +302,8 @@
...
+  "UPDATE_THUMBNAIL_LINK_TEXT": "Check Here"
```
pyphilia, line 308:
```suggestion
  "UPDATE_THUMBNAIL_LINK_TEXT": "Information"
```

pyphilia (changes requested):
Thanks for this PR! Removing all these tabs was really a good idea and removed bloating code, don't you think? Be careful of duplication, I think it can be easily avoided ;) Let me know if you have any questions.

Also UI reviews:
- Should we have the back button outlined? I think we don't see it enough. (we discuss it on slack)
- Clear chat button has a strange look now, was it all centered before?
Screenshot 2024-01-29 at 10 36 16

src/components/layout/Navigation.tsx
```diff
@@ -121,6 +128,7 @@ const Navigator = (): JSX.Element | null => {
       renderRoot={renderRoot}
       buildMenuItemId={buildNavigationLink}
       useChildren={useChildren as any}
```
spaenleh, line 130: I guess this was already from before, but is it possible to fix this `any` ? Or should we wait for the new version that hook anyway ?

src/components/pages/item/ItemPageLayout.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const enableEditing = permission
+    ? PERMISSIONS_EDITION_ALLOWED.includes(permission)
+    : false;
+
+  if (item && itemId && memberships) {
+    return (
+      <UppyContextProvider enable={enableEditing} itemId={itemId}>
+        <Box py={1} px={2}>
+          <Box display="flex" alignItems="center">
+            <BackButton onClick={() => navigate(buildItemPath(itemId))} />
```
spaenleh, line 52: Doesn't this create a "forward" navigation ? Maybe we could "rewind" the navigation ? But maybe that is a bit more involved ?

src/components/pages/item/ItemScreenLayout.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }, [itemId]);
+
+  const itemMembership = getHighestPermissionForMemberFromMemberships({
+    memberships,
+    memberId: currentMember?.id,
+  });
+  const permission = itemMembership?.permission;
+  const enableEditing = permission
+    ? PERMISSIONS_EDITION_ALLOWED.includes(permission)
+    : false;
```
spaenleh, line 38: This looks familiar from the ItemPageLayout component. Is it possible to access the upper Outlet context or do you need to redefine it ? if adding an Outlet makes you redefine the context maybe you can just do a passthrough of the values of the previous outlet ?

src/components/pages/item/type.ts
```diff
@@ -0,0 +1,6 @@
+import { DiscriminatedItem, PermissionLevel } from '@graasp/sdk';
+
+export interface OutletType {
+  item: DiscriminatedItem;
+  permission?: PermissionLevel;
```
spaenleh, line 5: Looking closely at this, I think that the optional can be avoided if the computations for the `permission` is moved inside the if in `ItemScreenLayout`. We should still ensure that access is allowed when public though. But this would simplify the logic in some places.

src/config/paths.ts
```diff
@@ -8,6 +8,16 @@ export const buildItemPath = (id = ':itemId'): string => `${ITEMS_PATH}/${id}`;
...
+export const ITEM_SETTINGS_PATH = 'settings';
+export const ITEM_INFORMATION_PATH = 'information';
+export const buildItemSettingsPath = (id = ':itemId'): string =>
+  `${ITEMS_PATH}/${id}/${ITEM_SETTINGS_PATH}`;
+export const buildItemSharePath = (id = ':itemId'): string =>
+  `${ITEMS_PATH}/${id}/${ITEM_SHARE_PATH}`;
+export const buildItemPublishPath = (id = ':itemId'): string =>
+  `${ITEMS_PATH}/${id}/${ITEM_PUBLISH_PATH}`;
+export const buildItemInformationPath = (id = ':itemId'): string =>
+  `${ITEMS_PATH}/${id}/${ITEM_INFORMATION_PATH}`;
```
spaenleh, line 23: I think react-router-dom has some methods to resolve these paths depending on where they are created. look at [`useResolvedPath`](https://reactrouter.com/en/main/hooks/use-resolved-path) which allows to use a relative path that is then resolved to the absolute path depending where you are in the app.

src/langs/constants.ts
```diff
@@ -366,4 +366,7 @@ export const BUILDER = {
   // temporary message
   "You can also find the items of this page in ''My Graasp''. This page will be unavailable soon.":
     "You can also find the items of this page in ''My Graasp''. This page will be unavailable soon.",
+  BACK: 'BACK',
```
spaenleh, line 369: I think the translations for the back button are not working as expected ... please check

spaenleh (approved):

src/components/pages/item/type.ts
pyphilia: The membership could be null, so the permission might not exist. I think you still want to proceed if the item is public, so well, permission can be `undefined`.

spaenleh: Okay sounds good

spaenleh (approved):
Nice !

spaenleh (approved):

PR #973, opened 2024-01-25
fix: hide non folder in tree modal

pyphilia:


pyphilia:
It does not work well with the pagination, so I close this PR.

PR #997, opened 2024-02-08
Use new graasp UI

spaenleh:


src/components/item/form/DocumentForm.tsx
```diff
@@ -44,29 +62,40 @@ export const DocumentExtraForm = ({
...
   );
 
+  const withFlavor = (textView: JSX.Element): JSX.Element =>
+    extra?.flavor ? (
+      <Alert severity={extra.flavor}>{textView}</Alert>
+    ) : (
+      textView
+    );
+
+  const handleChangeEditorMode = (mode: string) => {
```
ReidyT, line 79:
```suggestion
  const handleChangeEditorMode = (mode: EditorMode | `${EditorMode}`) => {
    setEditorMode(EditorMode[EditorMode[mode] as keyof typeof EditorMode]);
  }
```

Maybe this could solve as unknown but still have to cast.

ReidyT (approved):
Really nice ! LGTM 👍

src/components/item/form/DocumentForm.tsx
spaenleh: the issue is that the onChange method from the tabs does not return the enum type, so we can't make the typing work here. I opted to use a simple string instead. It is a bit ugly but in the end in JS it will not be safer.

package.json
```diff
@@ -10,23 +10,24 @@
...
-    "@graasp/sdk": "3.4.1",
-    "@graasp/translations": "1.22.1",
-    "@graasp/ui": "4.6.1",
-    "@mui/icons-material": "5.14.19",
-    "@mui/lab": "5.0.0-alpha.151",
-    "@mui/material": "5.14.19",
-    "@sentry/react": "7.93.0",
+    "@graasp/chatbox": "3.0.3",
+    "@graasp/query-client": "2.6.0",
+    "@graasp/sdk": "github:graasp/graasp-sdk#407-add-is-raw-on-document-extra",
```
pyphilia, line 24: Update!

src/components/Root.tsx
```diff
@@ -1,17 +1,20 @@
+import '@ag-grid-community/styles/ag-grid.min.css';
+import '@ag-grid-community/styles/ag-theme-material.min.css';
+import '@uppy/core/dist/style.css';
+// hack to have the text editor styles available without having to bundle them with graasp/ui
+import 'katex/dist/katex.min.css';
+import 'react-quill/dist/quill.snow.css';
+import 'react-toastify/dist/ReactToastify.css';
```
pyphilia, line 7: Was it better than building it in ui? Because now we have to do it for all platforms? or is it only necessary on edit?

src/components/item/form/DocumentForm.tsx
```diff
@@ -1,72 +1,101 @@
...
+  ITEM_FORM_DOCUMENT_TEXT_ID,
+} from '../../../config/selectors';
 import { BUILDER } from '../../../langs/constants';
 import { buildDocumentExtra } from '../../../utils/itemExtra';
 import type { EditModalContentPropType } from './EditModalWrapper';
 import NameForm from './NameForm';
 
+enum EditorMode {
+  Rich,
+  Raw,
```
pyphilia, line 36:
```suggestion
  Rich = 'rich',
  Raw = 'raw',
```

src/components/item/form/DocumentForm.tsx
```diff
@@ -77,24 +106,48 @@ export const DocumentExtraForm = ({
...
-          cancelButtonId={cancelButtonId}
-          showActions={showActions}
-        />
-      </Box>
-    </>
+      <TabContext value={editorMode.toString()}>
+        <Box sx={{ borderBottom: 1, borderColor: 'divider' }}>
+          <TabList
+            onChange={(_, value) => handleChangeEditorMode(value)}
+            aria-label="lab API tabs example"
```
pyphilia, line 113: Remove? Update?

src/components/item/form/DocumentForm.tsx
```diff
@@ -77,24 +106,48 @@ export const DocumentExtraForm = ({
...
+          >
+            <Tab
+              label={t(BUILDER.DOCUMENT_EDITOR_MODE_RICH_TEXT)}
+              value={EditorMode.Rich.toString()}
+            />
+            <Tab
+              label={t(BUILDER.DOCUMENT_EDITOR_MODE_RAW)}
+              value={EditorMode.Raw.toString()}
```
pyphilia, line 121: Is `toString()` always necessary?

pyphilia (approved):

src/components/Root.tsx
spaenleh: Building it in UI made a big CSS bundle with all of them, using them here allows to be a bit ore selective... 
I also was not able to make vite detect that there was the styles.css file available in the ui package ... it was failing with an error from the dependencies package ... apparently it is a bug on vite's side, so I avoided it completely this way. 

Also you only need to import the quill styles if you plan on displaying text from quill, same for katex.

src/components/item/form/DocumentForm.tsx
spaenleh: Since the tab switch only handles strings yes.

PR #1002, opened 2024-02-14
Fix update on unpublish

spaenleh:
See below is an extreme slow-mo of the button.

there is still an issue where while the fetching for the new itemPublish info is loading the state appears as not published... lets see if it is an issue or not.

https://github.com/graasp/graasp-builder/assets/39373170/c225cac7-8ee8-404e-afc1-39463fc88383


spaenleh:
@ReidyT I updated the code following your comment. The recording has also been updated. Let me know if it is better that way.

PR #1030, opened 2024-02-22
feat: update mode button, refactor

### LINKED ISSUE #1012: replace and use useCanUpdateItem
the hook can be changed to include `canEdit`, `canAdmin`, etc.

pyphilia:
- rename `ItemLayoutMode`
- Refactor `ModeButton`
- use `useGetPermissionForItem`, that will be replaced at some point, but at least it cleaned some old code
Screenshot 2024-02-22 at 16 43 11

close #1012

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -19,7 +19,6 @@ export const APP_VERSION = VITE_VERSION;
 export const DOMAIN = VITE_GRAASP_DOMAIN;
 
 export const API_HOST = VITE_GRAASP_API_HOST || 'http://localhost:3000';
-
 export const GRAASP_AUTH_HOST =
   VITE_GRAASP_AUTH_HOST || 'http://localhost:3001';
 export const GRAASP_PLAYER_HOST =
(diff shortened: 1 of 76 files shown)


LinaYahya (approved):

src/config/env.ts
```diff
@@ -1,30 +1,48 @@
...
+  VITE_GRAASP_AUTH_HOST,
+  VITE_SHOW_NOTIFICATIONS,
+  VITE_GRAASP_ASSETS_URL,
+  VITE_H5P_INTEGRATION_URL,
+  VITE_SENTRY_ENV,
+  VITE_GA_MEASUREMENT_ID,
+  VITE_SENTRY_DSN,
+} =
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
```
spaenleh, line 18: Mhmm you are just missing a `window.d.ts` file that exposes the Cypress key on the global object. We have examples of these in the apps and in auth I think.

spaenleh (approved):
Looks good !

PR #1031, opened 2024-02-23
feat: filter children in useChildren directly

ReidyT:
this is linked to graasp/graasp#831.


cypress/e2e/item/move/listMoveMultiple.cy.ts
```diff
@@ -74,13 +74,24 @@ describe('Move Items in List', () => {
...
+
+    //   itemIds.forEach((id) => {
+    //     cy.get(`${buildItemsTableRowIdAttribute(id)}`).should('not.exist');
+    //   });
+    // });
+
+    // Therefore, in this test, we can only verify that upon receiving a websocket message,
+    // the cache is invalidated and a refetch is triggered.
+
+    cy.wait('@moveItems');
```
pyphilia, line 94: I think you can still check that the requested body contains the correct ids.

pyphilia (approved):
LGTM! Maybe a slight change in the test so it's a little more complete 🛋️

cypress/e2e/item/move/listMoveMultiple.cy.ts
ReidyT: Yes you are right 😆!

Issue #1040, opened 2024-02-28
Add dropzone header


ztlee042:
https://www.figma.com/file/AMVdgMVt4RRW4j1YSmij5k/Drag-%26-Drop?type=design&node-id=2004%3A161&mode=design&t=4UWLuaCO4CshCqLt-1

PR #1042, opened 2024-02-29
1040 dropzone header

ztlee042:
#1040


src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+  const handleFiles = () => {
+    const { current } = ref;
+    if (current) {
+      const { files } = current;
+      if (files) {
+        // add files selected to uppy, this will upload them
+        [...files].map((file) =>
+          uppy?.addFile({ data: file, name: file.name }),
```
spaenleh, line 31: Add a comment specifying the need to add the `name` key in order to send the file name correctly.

src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+  return (
+    <Stack
+      justifyContent="flex-start"
+      alignItems="center"
+      direction="column"
+      flexGrow={1}
+      spacing={3}
+      sx={{ height: '100%', marginTop: '10%' }}
```
spaenleh, line 46: I see why you added a margin in percent, as centering does not look very good, but it has a bit of a strange behaviour when scaling the page horizontally.

src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+    >
+      <UploadFileOutlinedIcon sx={{ fontSize: '50px', color: '#5050d2' }} />
```
spaenleh, line 48: MUI Icons support specifying the use of colours from the palette:
```suggestion
      <UploadFileOutlinedIcon color="primary" sx={{ fontSize: '50px' }} />
```

Also we have the primary graasp colour string stored in a constant in case you need. But using the theme is the preffered approach.

src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <Typography variant="h4" sx={{ color: '#757575' }}>
```
spaenleh, line 49: The typography components also accept a `color` prop that can reference [the `theme.palette` properties](https://mui.com/material-ui/customization/default-theme/?expand-path=$.palette) using strings:
```suggestion
      <Typography variant="h4" color="text.secondary">
```

src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+        {t(BUILDER.DROPZONE_HELPER_TEXT)}
+      </Typography>
+      <Typography variant="h5" sx={{ color: '#757575' }}>
```
spaenleh, line 52: Same comment here.

src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+        {t('or')}
```
spaenleh, line 53: Use a translations string.

src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+        {t(BUILDER.DROPZONE_HELPER_ACTION)}
+      </Button>
+      <input
+        style={{ display: 'none' }}
+        type="file"
+        multiple
+        ref={ref}
+        onChange={handleFiles}
+      />
+      <Typography variant="h6" sx={{ color: '#757575' }}>
```
spaenleh, line 70: Not sure we need to have a heading6 here, maybe use `variant="body1"` to make it more uniform with the rest of the interface.

src/components/file/DropzoneHelper.tsx
```diff
@@ -0,0 +1,77 @@
...
+        {t(BUILDER.DROPZONE_HELPER_REMINDER)}
```
spaenleh, line 71:
```suggestion
        {t(BUILDER.DROPZONE_HELPER_LIMIT_REMINDER_TEXT)}
```

src/components/item/ItemMain.tsx
```diff
@@ -50,7 +51,7 @@ const ItemMain = ({ id, children, item }: Props): JSX.Element => {
   const { isChatboxMenuOpen, setIsChatboxMenuOpen } = useLayoutContext();
 
   return (
-    <Box id={id} m={2} className={ITEM_MAIN_CLASS}>
+    <Box id={id} m={2} className={ITEM_MAIN_CLASS} sx={{ height: '100%' }}>
```
spaenleh, line 54: The `Box` component supports most of the styling properties directly:
```suggestion
    <Box id={id} p={2} className={ITEM_MAIN_CLASS} height="100%">
```

I changed the `margin` to `padding` as it removes the tiny scrolling effect due to combination of the height 100% and the margin.

src/components/file/DropzoneHelper.tsx
spaenleh, line 1: A general consideration related to design, this is how the interface looks on mobile screens. Should the text be centered when wrapping ?
Screenshot 2024-03-04 at 08 31 01

spaenleh (changes requested):
Thank you for this PR. 🐨 

I added some comments related to how we usually style components. 

Please address the following before re-requesting a review:
- fix existing tests that are broken by the feature and should be fixed to match the new behaviour
- add tests that test the new feature 
- fix styling question on small screens

Also, this new component now appears in the following places: 
- Shared Items (if the user has no shared items)
- Favorite Items (if the user has no favorite items)
- Published Items (if the user has no published items)
- Trash (if the user has no items in the bin)
The message should not be displayed in those places. 

You should address this as we do not allow to upload files on the shared items screen for example. 

Let me know if you have questions or doubts.

cypress/e2e/item/upload/dropzoneUpload.cy.ts
```diff
@@ -0,0 +1,29 @@
+import { buildItemPath } from '@/config/paths';
+import { DROPZONE_HELPER_ID } from '@/config/selectors';
+
+import { SAMPLE_ITEMS, SAMPLE_PUBLIC_ITEMS } from '../../../fixtures/items';
+
+describe('home screen - dropzone helper visibility', () => {
```
spaenleh, line 6: for this I would wrap both tests in a describe that is called 'Dropzone Helper' and then have the two describe say 'Home Screen' & 'Empty Folder'.

spaenleh (approved):
Looks good!

PR #1050, opened 2024-03-04
feat: add checkbox in DescriptionForm to allow to add description above

ReidyT:
- Allow to define the placement of the description (above or below the item).
- See graasp/graasp-ui#694


pyphilia (changes requested):
Actually, could you have a select of enum instead of a checkbox? I think it also matches better your type for `descriptionPlacement` that is `Above` or `Below`. I don't think we will add left and right, but I think it's also clearer to the user if the default select is set on below.

pyphilia (changes requested):
Let's apply the changes as discussed on slack ;)

pyphilia (approved):

PR #1058, opened 2024-03-05
fix: revert previous published item layout

### LINKED ISSUE #1057: Fix published item icon in drawer menu

pyphilia:
This PR should be deployed on staging to prevent the bug to cascade on production.

fix #1057


spaenleh (approved):
Good catch !

PR #1086, opened 2024-03-20
fix: permission function

### LINKED ISSUE #1084: [bug] Highest permission is not right
This affects the permission on the apps, when you have `write` on the root and you create an app, the app receives `write` when it should be `admin`.

spaenleh:
In this PR:
- fix #1084
- add unit test step in ci tests


src/utils/item.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    expect(() => transformIdForPath(null)).toThrow();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    expect(() => transformIdForPath(undefined)).toThrow();
+  });
+
+  describe('getParentsIdsFromPath', () => {
```
pyphilia, line 40: I think this exists in sdk.

src/utils/item.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    expect(isUrlValid('somelink')).toBeFalsy();
+    expect(isUrlValid('graasp.eu')).toBeTruthy();
+    expect(isUrlValid('https://graasp')).toBeFalsy();
+
+    expect(isUrlValid('https://graasp.eu')).toBeTruthy();
+    expect(isUrlValid('http://graasp.eu')).toBeTruthy();
+    expect(isUrlValid('https://www.youtube.com/')).toBeTruthy();
+  });
+
+  it('transformIdForPath', () => {
```
pyphilia, line 25: This might exist in sdk.

pyphilia (approved):

src/utils/item.test.ts
spaenleh: Yes I will remove them in another PR, as during my first try a lot of tests broke, and I did not want to mix the bug fix and the refactoring.

PR #1094, opened 2024-03-21
fix: rename chatbox title

### LINKED ISSUE #986: Rename the chatbox drawer title to "Item name" chat
Currently the Drawer hosting the chatbox is named: `Chat`

We would like to rename it to: `"Item name" chat`

Do an item called `Math lesson` would generate a drawer with title: `"Math lesson" chat`

mariembencheikh:
fix #986

### DIFF
--- a/src/components/item/ItemMain.tsx
+++ b/src/components/item/ItemMain.tsx
@@ -63,7 +63,9 @@ const ItemMain = ({ id, children, item }: Props): JSX.Element => {
             direction="rtl"
           >
             <Typography variant="h6">
-              {translateBuilder(BUILDER.ITEM_CHATBOX_TITLE, {name: item.name})}
+              {translateBuilder(BUILDER.ITEM_CHATBOX_TITLE, {
+                name: item.name,
+              })}
             </Typography>
           </DrawerHeader>
           <Divider />
--- a/src/components/item/ItemMain.tsx
+++ b/src/components/item/ItemMain.tsx
@@ -6,7 +6,6 @@ import { DrawerHeader } from '@graasp/ui';
 import { RIGHT_MENU_WIDTH } from '../../config/constants';
 import { useBuilderTranslation } from '../../config/i18n';
 import { ITEM_MAIN_CLASS } from '../../config/selectors';
-import { BUILDER } from '../../langs/constants';
 import Chatbox from '../common/Chatbox';
 import { useLayoutContext } from '../context/LayoutContext';
 import ItemPanel from './ItemPanel';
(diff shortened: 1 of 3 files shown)


src/components/item/ItemMain.tsx
```diff
@@ -62,7 +61,7 @@ const ItemMain = ({ id, children, item }: Props): JSX.Element => {
             direction="rtl"
           >
             <Typography variant="h6">
-              {translateBuilder(BUILDER.ITEM_CHATBOX_TITLE)}
+              {translateBuilder(`"${item.name}"`)} chat
```
spaenleh, line 64: Is it correct that if my item is named `lesson 1` you are translating `lesson 1` and then concatenating it with ` chat` ?
I think the name of the item does not need to be translated. But the `chat` part does. For this you should use [i18next Interpolation feature](https://www.i18next.com/translation-function/interpolation). 
Some examples of this would be:
```ts
{translateBuilder(BUILDER.ITEM_CHATBOX_TITLE, { name: item.name })}
```
We will need to update `ITEM_CHATBOX_TITLE` string to be `"{{name}}" chat`.
It will then be translated in different languages: 
french: `Discussion pour "{{name}}"` for example.

spaenleh (changes requested):
Thank you for your first PR. 🎉 

I have added a comment. I think you should use interpolation. Let me know if you have questions or doubts.

src/langs/fr.json
```diff
@@ -79,7 +79,7 @@
   "INVITATIONS_TABLE_INVITATION_HEADER": "Invitation",
   "INVITATIONS_TABLE_PERMISSION_HEADER": "Permission",
   "ITEM_CATEGORIES_SELECTION_TITLE": "Catégories",
-  "ITEM_CHATBOX_TITLE": "Chatbox",
+  "ITEM_CHATBOX_TITLE": "Discussion pour",
```
spaenleh, line 82:
```suggestion
  "ITEM_CHATBOX_TITLE": "Discussion pour \"{{name}}\"",
```

src/components/item/ItemMain.tsx
```diff
@@ -61,7 +63,7 @@ const ItemMain = ({ id, children, item }: Props): JSX.Element => {
             direction="rtl"
           >
             <Typography variant="h6">
-              {translateBuilder(`"${item.name}"`)} chat
+              {translateBuilder(BUILDER.ITEM_CHATBOX_TITLE)} {`"${item.name}"`}
```
spaenleh, line 66:
```suggestion
              {translateBuilder(BUILDER.ITEM_CHATBOX_TITLE, {name: item.name})}
```

spaenleh (changes requested):
You should use the interpolation from the i18next library. See the link in my last comment.

spaenleh (approved):
Looks good to me 🎄

PR #1112, opened 2024-03-25
feat: add geolocation modal

pyphilia:
Screenshot 2024-03-25 at 17 08 21


src/components/item/settings/GeolocationModalButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const { t } = useBuilderTranslation();
+  const { t: commonT } = useCommonTranslation();
+  const [open, setOpen] = useState(false);
+  const { data: geoloc } = hooks.useItemGeolocation(item.id);
+  const { mutate: saveGeoloc } = mutations.usePutItemGeolocation();
+
+  const helperLabelRef = useRef<HTMLInputElement>(null);
+  const addressLabelRef = useRef<HTMLInputElement>(null);
+  const latRef = useRef<HTMLInputElement>(null);
+  const lngRef = useRef<HTMLInputElement>(null);
```
spaenleh, line 37: Any reason why we use refs and not controlled inputs ?

src/langs/fr.json
```diff
@@ -340,5 +340,15 @@
   "ITEM_SETTINGS_IS_COLLAPSED_TITLE": "Minifier un élément",
   "ITEM_SETTINGS_DESCRIPTION_PLACEMENT_ABOVE": "Au-dessus du contenu",
   "ITEM_SETTINGS_DESCRIPTION_PLACEMENT_BELOW": "En-dessous du contenu",
-  "ITEM_SETTINGS_DESCRIPTION_PLACEMENT_TITLE": "Placement de la description"
+  "ITEM_SETTINGS_DESCRIPTION_PLACEMENT_TITLE": "Placement de la description",
+  "ITEM_GEOLOCATION_ADVANCED_BUTTON": "Localisation avancée",
+  "ITEM_GEOLOCATION_ADVANCED_MODAL_TITLE": "Paramètres avancés pour la localisation",
+  "ITEM_GEOLOCATION_ADVANCED_MODAL_DESCRIPTION": "Aucune information soumise par ce formulaire ne sera validée. L'exactitude des données revient à l'éditeur.",
```
spaenleh, line 346: "Les informations fournies via ce formulaire ne sont pas validées. Vérifiez l'exactitude des données."

spaenleh (approved):
Thank you for this PR.
Would it make sens to put the country dropdown before the address ? so we can go from less specific to most specific ? 

Can users input only lat+long or the address ? How can we make it clear that it is an OR choice and that the other parts will get autofilled (if that is the case) ?

spaenleh (approved):
Thank you for this PR.
Would it make sens to put the country dropdown before the address ? so we can go from less specific to most specific ? 

Can users input only lat+long or the address ? How can we make it clear that it is an OR choice and that the other parts will get autofilled (if that is the case) ?

spaenleh (approved):
Thank you for this PR.
Would it make sens to put the country dropdown before the address ? so we can go from less specific to most specific ? 

Can users input only lat+long or the address ? How can we make it clear that it is an OR choice and that the other parts will get autofilled (if that is the case) ?

src/components/item/settings/GeolocationModalButton.tsx
pyphilia: Isn't it best for performance? If I use controlled buttons I also need `useEffect` on values.

PR #1130, opened 2024-04-04
feat: update map for mobile use

pyphilia:
This PR adds a new route `/map` that will display the map "in full screen". This is used by the mobile app to display the map.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -83,11 +83,10 @@ const App = (): JSX.Element => {
     RecycledItemsScreen,
     withAuthorizationProps,
   );
-  const MapWithAuthorization = MapItemsScreen;
 
   return (
     <Routes>
-      <Route path={MAP_ITEMS_PATH} element={<MapWithAuthorization />} />
+      <Route path={MAP_ITEMS_PATH} element={<MapItemsScreen />} />
       <Route
         element={
           <Main>
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -14,6 +14,7 @@ import {
   ITEM_PUBLISH_PATH,
   ITEM_SETTINGS_PATH,
   ITEM_SHARE_PATH,
+  MAP_ITEMS_PATH,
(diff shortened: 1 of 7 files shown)


src/components/item/MapView.tsx
```diff
@@ -9,18 +9,21 @@ import { buildPlayerTabName } from '@/config/selectors';
 
 type Props = {
   parentId?: DiscriminatedItem['id'];
-  title: string;
+  title?: string;
+  height?: string;
 };
 
-const MapView = ({ parentId, title }: Props): JSX.Element => {
+const MapView = ({ parentId, title, height = '80vh' }: Props): JSX.Element => {
```
spaenleh, line 16: The container is a Flex item, so it should be possible to grow it to fill the page.

src/config/paths.ts
```diff
@@ -12,6 +12,7 @@ export const ITEM_ID_PARAMS = 'itemId';
 export const ITEM_SHARE_PATH = 'share';
 export const ITEM_PUBLISH_PATH = 'publish';
 export const ITEM_SETTINGS_PATH = 'settings';
+export const MAP_ITEMS_PATH = 'map';
```
spaenleh, line 15: So it will be `https://builder.graasp.org/map?parentId=1234` ?

src/components/App.tsx
```diff
@@ -81,9 +83,11 @@ const App = (): JSX.Element => {
     RecycledItemsScreen,
     withAuthorizationProps,
   );
+  const MapWithAuthorization = MapItemsScreen;
```
spaenleh, line 86: So no auth currently ?

spaenleh (approved):
Looks good, just some questions

src/components/App.tsx
pyphilia: you can access the map without auth, if the item is public.

ReidyT (approved):
Looks good 👍

PR #1140, opened 2024-04-09
feat: use packed item

pyphilia:
- remove most of the memberships calls
- remove most of the get tags calls
- update most tests: avoid using same fixtures in all files
- added back some timeout... I think here the loading is way quicker so it has the time to render a second time. My guess is that it will be resolved when we refactor the table.


ReidyT (approved):
Thanks for this amazing PR ! LGTM 😄

cypress/e2e/item/share/itemLogin.cy.ts
```diff
@@ -1,6 +1,11 @@
-import { ItemLoginSchemaType } from '@graasp/sdk';
+import {
+  ItemLoginSchemaType,
+  PackedFolderItemFactory,
+  PackedItem,
+  PermissionLevel,
+} from '@graasp/sdk';
 
-import { v4 as uuidv4 } from 'uuid';
+import { v4 as uuidv4, v4 } from 'uuid';
```
spaenleh, line 8: Mhmmm ?

spaenleh (approved):
Looks good, just look out for other `.only`

Issue #1141, opened 2024-04-10
fix: display interpolation data for chatbox name


mariembencheikh:
an issue related to display interpolation data on mouse hover for chatbox name
#986

PR #1142, opened 2024-04-10
fix: display interpolation data on mouse hover for chatbox name

mariembencheikh:
in this PR fix an issue related to display interpolation data on mouse hover for chatbox name
#1141


spaenleh (approved):
Thank you for the fix 👍

PR #1149, opened 2024-04-16
fix: remove double wraping of main

spaenleh:


ReidyT (approved):
Nice ! It's directly cleaner 😄

PR #1164, opened 2024-04-18
feat: remove country form, update map

### LINKED ISSUE #1161: [geoloc] improve country field
Use https://www.npmjs.com/package/i18n-iso-countries
- show country name depending on language
- translate country string to code, show error if country does not exist

pyphilia:
In the end I disabled the country form, since it is not used by the backend (country is automatically computed in the backend). And the user will write the country in the address probably.

close #1161


src/components/item/settings/GeolocationModalButton.tsx
```diff
@@ -35,7 +34,7 @@ export const GeolocationModalButton = ({ item }: Props): JSX.Element => {
   const addressLabelRef = useRef<HTMLInputElement>(null);
   const latRef = useRef<HTMLInputElement>(null);
   const lngRef = useRef<HTMLInputElement>(null);
-  const [country, setCountry] = useState<string>();
+  // const [country, setCountry] = useState<string>();
```
spaenleh, line 37: remove ?

spaenleh (approved):

Issue #1162, opened 2024-04-17
Redesign Add Link Interface


ztlee042:
![image](https://github.com/graasp/graasp-builder/assets/31677780/e94a0f86-4184-4843-b612-e853ce682864)

PR #1165, opened 2024-04-18
feat: redesign add link interface

ztlee042:
#1162

In this PR:

- Display name form is slightly modified to display a different message when used in LinkForm
- LinkForm is improved to provide more information
- Two preview components are added for users to choose the desired display type they want


cypress/support/viewUtils.ts
```diff
@@ -108,7 +108,8 @@ export const expectLinkViewScreenLayout = ({
   }
 
   if (!html && (settings?.showLinkButton ?? DEFAULT_LINK_SHOW_BUTTON)) {
-    cy.get('[data-testid="OpenInNewIcon"]').should('be.visible');
+    // this data-testid is set in graasp/ui
+    cy.get('[data-testid="fancy-link-card"]').should('be.visible');
```
ReidyT, line 112: Is the UI exposed a constant for that ?

package.json
```diff
@@ -21,10 +21,10 @@
     "@emotion/styled": "11.11.5",
     "@graasp/chatbox": "3.1.0",
     "@graasp/map": "1.12.1",
-    "@graasp/query-client": "3.8.0",
-    "@graasp/sdk": "4.11.0",
+    "@graasp/query-client": "3.9.0",
+    "@graasp/sdk": "4.12.0",
     "@graasp/translations": "1.28.0",
-    "@graasp/ui": "4.18.2",
+    "@graasp/ui": "github:graasp/graasp-ui#835-fancy-link-ui",
```
ReidyT, line 27: just a reminder 😆

src/components/item/form/link/LinkForm.tsx
```diff
@@ -0,0 +1,393 @@
...
+        setChanges={onChange}
+        updatedProperties={updatedProperties}
+      />
+      <TextField
+        label="Description"
+        variant="standard"
+        InputLabelProps={{ shrink: true }}
+        value={updatedProperties.description}
+        onChange={onChangeDescription}
+        // helperText={
```
ReidyT, line 260: remove ?

src/components/item/form/link/LinkForm.tsx
```diff
@@ -0,0 +1,393 @@
...
+                  }
+                  control={<Radio />}
+                  slotProps={{
+                    typography: {
+                      width: '100%',
+                      minWidth: '0px',
+                    },
+                  }}
+                  // sx={{
+                  //   // this ensure the iframe takes up all horizontal space
```
ReidyT, line 374: remove

src/components/item/form/link/LinkForm.tsx
ReidyT, line 1: This component is a bit long, if it is possible it could be great to break it in smaller components and/or extracting some logic in hooks or utils functions.

src/components/item/form/link/LinkForm.tsx
```diff
@@ -0,0 +1,393 @@
...
+                  }}
+                />
+              )}
+              {linkData?.html && linkData.html !== '' && (
+                <StyledFormControlLabel
+                  value={LinkType.Embedded}
+                  label={
+                    // eslint-disable-next-line react/no-danger
+                    <StyledDiv
+                      sx={{}}
```
ReidyT, line 362: it is necessary to have an empty sx ? If yes, maybe we should make it optional in the StyledDiv.

ReidyT (changes requested):
Thanks for this PR ! I think we could remove unused props and maybe we can try to refactor the LinkForm to have smaller components or extracting some logic ? What do you think ?

cypress/support/viewUtils.ts
spaenleh: UI has a constant for that, but I am not sure we should use it in the test because:
- It is possible that cypress will try to bundle all of ui for the tests even if we only use a constant
- I am not sure we want to depend on the constant, I think there should be a better way to test the presence of the card ... 

@ReidyT Is it ok if we keep it like this ?

ReidyT: Thanks for your response. Yes, I think you can let it like that, especially if we risk to bundle all UI just for this... It was more a question 😄. Normally this constant in UI shouldn't changed frequently so it is ok.

src/components/item/form/link/LinkForm.tsx
spaenleh: Thank you !

spaenleh: I did some refactoring and added utility functions with thests.

ReidyT (approved):
Thanks for this refactor 😆, LGTM 🚀🙌

PR #1188, opened 2024-04-25
feat: redesign add application ui

ztlee042:
In this PR:

- The add application UI has been changed
- A "mostly used" section has been added (not ready yet).
- A "+ Create your own app" button is added
- Change the width of NameForm and DisplayForm to fullwidth

https://github.com/graasp/graasp-builder/assets/31677780/5c8fd4ef-bd7c-4eb4-88ef-1f4ec881aef2


src/langs/en.json
```diff
@@ -198,7 +198,7 @@
   "NAVIGATION_PUBLISHED_ITEMS_TITLE": "Published Items",
   "NAVIGATION_MY_ITEMS_TITLE": "My Graasp",
   "NAVIGATION_SHARED_ITEMS_TITLE": "Shared Items",
-  "NEW_ITEM_APP_TAB_TEXT": "Application",
+  "NEW_ITEM_APP_TAB_TEXT": "Add Application",
```
pyphilia, line 201:
```suggestion
  "NEW_ITEM_APP_TAB_TEXT": "Add an Application",
```

src/langs/en.json
```diff
@@ -292,8 +292,14 @@
   "DELETE_LAST_ADMIN_ALERT_MESSAGE": "You are not allowed to delete this admin as this is the only admin",
   "APPROVE_BUTTON_TEXT": "OK",
   "APP_URL": "App Url",
+  "APP_SECTION_FULL_LIST": "Full List",
```
pyphilia, line 295: I suggest

```suggestion
  "APP_SECTION_FULL_LIST": "All applications",
```

src/langs/en.json
```diff
@@ -292,8 +292,14 @@
...
+  "APP_SECTION_MOSTLY_USED": "Mostly Used",
```
pyphilia, line 296:
```suggestion
  "APP_SECTION_MOSTLY_USED": "Most Used",
```

src/langs/en.json
```diff
@@ -292,8 +292,14 @@
...
+  "APP_SECTION_FULL_LIST_BUTTON_TEXT": "Show full list",
```
pyphilia, line 297:
```suggestion
  "APP_SECTION_FULL_LIST_BUTTON_TEXT": "Show all applications",
```

src/langs/en.json
```diff
@@ -292,8 +292,14 @@
...
+  "APP_SECTION_MOSTLY_USED_BUTTON_TEXT": "Show mostly used",
```
pyphilia, line 298:
```suggestion
  "APP_SECTION_MOSTLY_USED_BUTTON_TEXT": "Show most used",
```

src/langs/en.json
```diff
@@ -292,8 +292,14 @@
...
+  "ADD_BUILT_IN_APP_TEXT": "Search the app here...",
```
pyphilia, line 299:
```suggestion
  "ADD_BUILT_IN_APP_TEXT": "Search the application here...",
```

src/langs/en.json
```diff
@@ -292,8 +292,14 @@
...
   "CREATE_CUSTOM_APP": "Add Your Custom App",
-  "CREATE_CUSTOM_APP_DESCRIPTION": "Advanced option to create a custom app",
+  "CREATE_CUSTOM_APP_LABEL": "Need a custom solution?",
+  "CREATE_CUSTOM_APP_DESCRIPTION": "Create your own app",
```
pyphilia, line 302:
```suggestion
  "CREATE_CUSTOM_APP_DESCRIPTION": "Create your own application",
```

src/components/item/form/AppForm.tsx
```diff
@@ -14,26 +13,49 @@ import { sortByName } from '@/utils/item';
...
 }: AppGridProps): JSX.Element | JSX.Element[] => {
-  const { useApps } = hooks;
+  const { useApps, useMostUsedApps } = hooks;
   const { data, isLoading } = useApps();
-
+  const { data: dataMostUsed } = useMostUsedApps();
   const { t: translateBuilder } = useBuilderTranslation();
 
-  if (data) {
+  if (dataMostUsed && showFull === false) {
```
pyphilia, line 36:
```suggestion
  if (dataMostUsed && !showFull) {
```

src/components/item/form/AppForm.tsx
```diff
@@ -14,26 +13,49 @@ import { sortByName } from '@/utils/item';
...
+    return (
+      <>
```
pyphilia, line 38: I think you don't need this.

src/components/item/form/AppForm.tsx
```diff
@@ -14,26 +13,49 @@ import { sortByName } from '@/utils/item';
...
+                handleSelection({ url: ele.url, name: ele.name });
+              }
+            }}
+          />
+        ))}
+      </>
+    );
+  }
+
+  if (data && showFull === true) {
```
pyphilia, line 58:
```suggestion
  if (data && showFull) {
```

src/components/main/AppCard.tsx
```diff
@@ -39,77 +41,58 @@ const AppCard = ({
...
-              variant="body2"
-              color="text.secondary"
-              width="100%"
+  <Tooltip title={description || ''}>
+    <Card
+      sx={{
+        width: '100%',
+        maxHeight: '80px',
+        flex: '1 1 calc(50% - 8px)',
+        outline: selected ? '2px solid #5050d2' : '',
```
pyphilia, line 50: Use theme for the color.
Something like `theme.palette.primary.main` where `theme = useTheme`

src/components/main/AppCard.tsx
```diff
@@ -39,77 +41,58 @@ const AppCard = ({
...
+              overflow="hidden"
+              justifyContent="center"
+            />
+          </Grid>
+          <Grid xs={10.5}>
+            <CardContent
               sx={{
-                display: { xs: 'none', sm: 'block', md: 'block' },
+                pt: 1,
+                paddingBottom: '0 !important',
```
pyphilia, line 74: Was the `important` necessary?

pyphilia (changes requested):
Thanks for those changes! I think it will really improve the creation of apps!

Some additional design comments:
I have those "margins", I think because of overflowing.

Screenshot 2024-05-01 at 11 43 26
Screenshot 2024-05-01 at 11 43 19

Also I know were not really responsive friendly, but I think it would be good to at least collapse the 2 columns into one on smaller screens. 
Screenshot 2024-05-01 at 12 06 15

ztlee042:
Hi Kim @pyphilia, I have updated the styles. Could you please check if the margin problems are still there? Thanks a lot.

src/components/main/AppCard.tsx
ztlee042: I am afraid so, the Card Content component has a built-in weird bottom padding that is lager than the other dimensions. The `important` was used to override this behavior.

ztlee042:
I just had two feedback from the testing participants:

- The search bar could stand out more. Both of them didn't notice the search bar at first look.
- The scroll bar should be visible. They didn't realize it was a scrollable area at first look.

src/components/item/form/AppForm.tsx
```diff
@@ -168,39 +196,71 @@ const AppForm = ({ onChange, updatedProperties = {} }: Props): JSX.Element => {
...
+                ? translateBuilder(BUILDER.APP_SECTION_ALL_APP_BUTTON_TEXT)
+                : translateBuilder(BUILDER.APP_SECTION_MOST_USED_BUTTON_TEXT))}
+          </Button>
+        </Stack>
+        <Stack
+          flexDirection="row"
+          flexWrap="wrap"
+          width="100%"
+          gap={1}
+          height="300px"
```
pyphilia, line 239: You should try to get rid of this random value. Probably a working idea is to include to rest of the form in the `Stack` as well. You can also set some `minHeight` or `maxHeight`.

src/components/item/form/AppForm.tsx
```diff
@@ -168,39 +196,71 @@ const AppForm = ({ onChange, updatedProperties = {} }: Props): JSX.Element => {
...
+        </Stack>
+      </Stack>
+      <Typography variant="body2">
+        {translateBuilder(BUILDER.CREATE_CUSTOM_APP_LABEL)}
+      </Typography>
+      <Button
+        startIcon={<Add fontSize="small" />}
+        variant="contained"
+        onClick={addCustomApp}
+        id={CUSTOM_APP_CYPRESS_ID}
```
pyphilia, line 260:
```suggestion
        id={CUSTOM_APP_BUTTON_ID}
```

src/components/main/AppCard.tsx
```diff
@@ -38,78 +41,61 @@ const AppCard = ({
...
-        width="100%"
-        spacing={[0, 1]}
+}: Props): JSX.Element => {
+  const theme = useTheme();
+  return (
+    <Tooltip title={description || ''}>
+      <Card
+        sx={{
+          minWidth: '180px',
+          maxHeight: '80px',
```
pyphilia, line 51: Do you really need those values?

src/components/main/AppCard.tsx
```diff
@@ -38,78 +41,61 @@ const AppCard = ({
...
-  </Grid2>
-);
+        <StyledCardActionArea>
+          <Grid container direction="row" alignItems="center">
+            <Grid xs={1.5} display="flex" justifyContent="center">
+              <Box
+                component="img"
+                src={image ?? defaultImage}
+                alt={name}
+                width="80%"
```
pyphilia, line 65: Why? I think you can use `100%`. If you need margin, you should add `spacing` to `Grid`. Probably doing so will allow you to use an integer for `xs` as well.

pyphilia (changes requested):
Thanks for the work!

Some additional thoughts:
- "the scrollbar should be visible" -> I think it's always visible, do you mean make it even more visible? Wouldn't that be too much?
- Also I think the "create your own application" button is way more too visible, that's the very first thing I see in this modal. Actually it doesn't have to be that highlighted, why not just a little text next to "Need a custom solution ? <click here to embed your own app>" for example.

ztlee042:
> Thanks for the work!
> 
> Some additional thoughts:
> 
> * "the scrollbar should be visible" -> I think it's always visible, do you mean make it even more visible? Wouldn't that be too much?
> * Also I think the "create your own application" button is way more too visible, that's the very first thing I see in this modal. Actually it doesn't have to be that highlighted, why not just a little text next to "Need a custom solution ? " for example.

For the first question, on Mac [there is a setting set scrollbar invisible by default](https://www.macrumors.com/how-to/make-scroll-bars-always-visible/), that's why users didn't see the scrollbar by default.
I will look into the second question.

Thanks for your feedback overall!

PR #1211, opened 2024-05-02
fix: remove navigation root and own and shared calls

pyphilia:
Screenshot 2024-05-02 at 14 50 20

I removed the root menu, because it is a pb to use own and shared endpoints. The first arrow is not clickable anymore.


src/components/layout/Navigation.tsx
```diff
@@ -42,69 +36,29 @@ const Navigator = (): JSX.Element | null => {
...
-        <ItemMenu
-          itemId="root"
-          useChildren={
-            isParentOwned || pathname === HOME_PATH
-              ? (useOwnItems as any)
-              : useSharedItems
-          }
-          buildToItemPath={buildToItemPath}
-        />
+        <IconButton id={NAVIGATION_HOME_ID} onClick={() => navigate(HOME_PATH)}>
```
spaenleh, line 53: I like that the home is now clickable, but maybe we should use a react-router link passed in the component prop instead of a js navigate ? This would allow opening in a new tab etc.

spaenleh (approved):
Changes look good ! Maybe a small nitpick on using Link instead of useNavigate for the home button.

PR #1234, opened 2024-05-13
feat: update map, pass parent item

pyphilia:
Update map, and pass full item parent to the map. Needed to get the permission over the parent.


spaenleh (approved):
👍

PR #1246, opened 2024-05-22
feat: add query param for mode

### LINKED ISSUE #1245: add query param for view
This will allow to show the map directly, given a link. And also allow to come back from player for example.

pyphilia:
Handle query param to keep track of the view mode.
It also simplifies tests, so I might change all of them in a next PR.

close #1245


src/components/item/sharing/shortLink/ShortLinksRenderer.tsx
```diff
@@ -59,6 +61,11 @@ const ShortLinksRenderer = ({
       const clientHostManager = ClientHostManager.getInstance();
       const url = clientHostManager.getItemAsURL(platform, itemId);
 
+      // not ideal
+      if (platform === Context.Builder) {
+        url.searchParams.set('mode', mode);
+      }
```
spaenleh, line 67: This implies that the mode used in the link will be the mode currently used. Do we want to allow users to select the mode they want to share ? temporarily this would be enough, but we should think about these design questions.

src/components/pages/item/ItemPageLayout.tsx
```diff
@@ -27,7 +29,14 @@ const ItemPageLayout = (): JSX.Element => {
...
         <Box display="flex" alignItems="center">
-          <BackButton onClick={() => navigate(buildItemPath(itemId))} />
+          <BackButton
+            onClick={() =>
+              navigate({
+                pathname: buildItemPath(itemId),
+                search: search.toString(),
+              })
+            }
+          />
```
spaenleh, line 39: Could this be a `Link` instead of a JS onClick ?

src/config/externalPaths.ts
```diff
@@ -9,8 +11,14 @@ import { buildItemPath } from './paths';
 
 export const buildGraaspPlayerView = (id: string): string =>
   `${GRAASP_PLAYER_HOST}/${id}`;
-export const buildGraaspBuilderView = (id: string): string =>
-  `${window.location.origin}${buildItemPath(id)}`;
+export const buildGraaspBuilderView = (
+  id: string,
+  mode = DEFAULT_ITEM_LAYOUT_MODE,
+): string => {
+  const url = new URL(`${window.location.origin}${buildItemPath(id)}`);
```
spaenleh, line 18: You can make use of the `base` argument of `URL`:
```suggestion
  const url = new URL(buildItemPath(id), window.location.origin);
```

src/config/paths.ts
```diff
@@ -1,9 +1,20 @@
...
 export const ITEMS_PATH = '/items';
-export const buildItemPath = (id = ':itemId'): string => `${ITEMS_PATH}/${id}`;
+export const buildItemPath = (
+  id = ':itemId',
+  args: { mode?: ItemLayoutMode } = {},
+): string => {
+  let qs = '';
+  if (args.mode) {
+    qs = `?mode=${args.mode}`;
+  }
```
spaenleh, line 15: I think you can use `URLSearchParams` to build it:
```suggestion
  const search = new URLSearchParams();
  if (args.mode) {
    search.set("mode", args.mode);
  }
  // use then search.toString()
```

spaenleh (approved):
Looks good, I suggest using URLSearchParams and builders when possible. 

I like where this is going !

PR #1255, opened 2024-05-29
fix: embedding setting and iframe display

spaenleh:
In this PR I fix an issue where the embedding option would not be shown because the html prop in not returned when no reiche media embedding is possible.
I also fixed the iframe sizing issue (by using the ui PR)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@graasp/query-client": "3.9.0",
     "@graasp/sdk": "4.12.0",
     "@graasp/translations": "1.28.0",
-    "@graasp/ui": "github:graasp/graasp-ui#iframe-div-issue",
+    "@graasp/ui": "4.19.1",
     "@mui/icons-material": "5.15.18",
     "@mui/lab": "5.0.0-alpha.170",
     "@mui/material": "5.15.18",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "@graasp/query-client": "3.9.0",
     "@graasp/sdk": "4.12.0",
     "@graasp/translations": "1.28.0",
-    "@graasp/ui": "4.19.0",
+    "@graasp/ui": "github:graasp/graasp-ui#iframe-div-issue",
```
pyphilia, line 27: Update.

src/components/item/form/link/LinkForm.tsx
```diff
@@ -244,25 +244,26 @@ const LinkForm = ({
...
-                    sx={{
-                      // this ensure the iframe takes up all horizontal space
-                      '& iframe': {
-                        width: '100%',
-                      },
-                    }}
-                  />
-                )
+                linkData?.isEmbeddingAllowed &&
+                  (linkData?.html === undefined || linkData?.html === '') && (
```
pyphilia, line 248: Is that not enough?
```suggestion
                  (!linkData?.html) && (
```

pyphilia (approved):

src/components/item/form/link/LinkForm.tsx
spaenleh: Yes indeed !

PR #1266, opened 2024-06-06
fix: link default to use showLinkButton = false

spaenleh:
The default link type is a bare link, but when not specified the backend creates a link with the showLinkButton to true. The default setting was added in the frontend to give the expected result.


ReidyT (approved):
LGTM 👍

pyphilia (approved):

PR #1284, opened 2024-06-11
feat: don't allow none folder items to be published

ReidyT:
- also improve the description components of the status button

Screenshot 2024-06-11 at 15 37 52
Screenshot 2024-06-11 at 15 37 31
Screenshot 2024-06-11 at 15 37 15
Screenshot 2024-06-11 at 15 36 51
Screenshot 2024-06-11 at 15 36 30
Screenshot 2024-06-11 at 15 34 41
Screenshot 2024-06-11 at 15 33 28
Screenshot 2024-06-11 at 15 31 36
image


cypress/e2e/item/publish/publishedItem.cy.ts
```diff
@@ -340,4 +341,20 @@ describe('Public Item', () => {
       waitOnUnpublishItem(publicItem);
     });
   });
+
+  describe('Cannot Publish None Folder Item', () => {
```
pyphilia, line 345:
```suggestion
  describe('Cannot Publish Non-Folder Item', () => {
```

src/components/hooks/usePublicationStatus.tsx
```diff
@@ -83,6 +84,8 @@ const computePublicationStatus = ({
   switch (true) {
     case isPublishedChildren({ item, publishedEntry }):
       return PublicationStatus.PublishedChildren;
+    case itemTypeNotAllowedToBePublished(item.type):
```
pyphilia, line 87: Same remark as in the backend. You can create your own function for the frontend.

src/langs/en.json
```diff
@@ -186,6 +186,7 @@
   "LIBRARY_SETTINGS_PUBLISH_NOTIFICATIONS_LABEL": "Send email notifications to all co-editors",
   "LIBRARY_SETTINGS_PUBLISHED_STATUS": "This element is published. Anyone can access it and is available on Graasp Library, our public repository of learning ressources.",
   "LIBRARY_SETTINGS_CHILD_PUBLISHED_STATUS": "This element is part of a collection that is already published. Publishing and un-publishing can be done at the collection root.",
+  "LIBRARY_SETTINGS_TYPE_NOT_ALLOWED_STATUS": "This element can not be published. Only folders are allowed to be published on Graasp Library.",
```
pyphilia, line 189: Should we have dynamic text for types? I think we have translations for types as well in common.

src/langs/en.json
```diff
@@ -218,6 +219,7 @@
   "LIBRARY_SETTINGS_PUBLICATION_STATUS_INVALID": "Invalid",
   "LIBRARY_SETTINGS_PUBLICATION_STATUS_OUTDATED": "Outdated",
   "LIBRARY_SETTINGS_PUBLICATION_STATUS_UNPUBLISHED": "Unpublished",
+  "LIBRARY_SETTINGS_PUBLICATION_STATUS_TYPE_NOT_ALLOWED": "Item Type not allowed",
```
pyphilia, line 222: I would have something more specific (file cannot be published, link cannot be published)

pyphilia (changes requested):
Thanks for the changes! 
Actually I wonder if it wouldn't be better to hide the publication page completely for given types? Or just show a text, instead of providing the form on the left but it's unnecessary to fill?

src/langs/en.json
ReidyT: Really good idea !

src/langs/en.json
ReidyT: Yes it's an interesting point !

src/components/hooks/usePublicationStatus.tsx
ReidyT: As I wrote in the backend PR, for me it is not an issue to use a function from the SDK, maybe we can discuss in the backend's PR ?

pyphilia (approved):

Issue #572, opened 2023-02-22
Improve file upload dialog


spaenleh:
File upload dialog can be improved to allow better user feedback.

- [x] "Done" button should either be removed OR close the dialog to let users know that the operation was performed successfully.
- [ ] Height of the file upload dialog should be adapted to the height of the modal.
- [x] #846

PR #1295, opened 2024-06-21
feat: remove uppy and use upload hooks

pyphilia:
This PR removes the use of uppy everywhere. We now use real mutations to upload files for:
- item thumbnail: I didn't change the layout
- h5p: a button
- zip: a button
- files: file drop zone

My implementation is a mixed of showing upload progress within the drop zone and toaster. The reason is that we sometimes clear the dropzone before the complete upload is finished. Here are the 3 ways we can upload files:
- from the dropzone in the screen (when the folder is empty): use toast because the view is updated at the first item upload (websockets). 
- from new item modal: the loading is show in the dropzone, and the modal closes when it finishes
- drop in a folder that already contains items: the dropzone disappear immediately, so the loading and errors happens in toastr.

A bug I've noticed is that the backend actually cannot handle multiple files upload (it gets stuck and never return for more than 1 file). Uploading one file is okay, and was actually working with uppy because uppy was sending one file at a time! So I faked this behavior by sending one request per file. 
Something that could considered a glitch is "showing n notifications when uploading n files". This is because the notifications are triggered by query client. Personally I think it is okay, but let me know what you think.

Screenshot 2024-06-21 at 17 56 23

Screenshot 2024-06-21 at 17 44 32

https://github.com/graasp/graasp-builder/assets/11229627/647dd533-962d-40d5-897c-105191c0f5d8

TODO:
- [ ] write tests


src/components/hooks/uploadWithProgress.tsx
```diff
@@ -0,0 +1,51 @@
+import { useRef } from 'react';
+import { Id, toast } from 'react-toastify';
+
+import { AxiosProgressEvent } from 'axios';
+
+import { useBuilderTranslation } from '@/config/i18n';
+
+// eslint-disable-next-line import/prefer-default-export
```
spaenleh, line 8: Maybe we can also disable this rule.

src/components/hooks/uploadWithProgress.tsx
```diff
@@ -0,0 +1,51 @@
...
+    if (toastId.current) {
+      toast.done(toastId.current);
+      toastId.current = null;
+    }
+  };
+  const closeAndShowError = (e: Error) => {
+    toast.error(e.message);
+    if (toastId.current) {
+      toastId.current = null;
+    }
```
spaenleh, line 47: You do not close the toast with `toastId` in this case ?

spaenleh (approved):
Looks good, just a few nitpicks and comments ! Very nice !

spaenleh:
@pyphilia Does this PR also fix some of the issues highlighted in #572 ?

PR #1347, opened 2024-07-15
fix: show link icon in table

pyphilia:
Shows link thumbnail or icon if it exists in the list view.


LinaYahya (approved):

package.json
```diff
@@ -22,7 +22,7 @@
     "@graasp/chatbox": "3.1.0",
     "@graasp/map": "1.16.0",
     "@graasp/query-client": "3.15.2",
-    "@graasp/sdk": "4.17.0",
+    "@graasp/sdk": "github:graasp/graasp-sdk#link-thumbnail",
```
ReidyT, line 25: just a reminder :)

src/components/item/form/link/LinkForm.tsx
```diff
@@ -213,7 +213,9 @@ const LinkForm = ({
                   <LinkCard
                     title={linkData?.title || ''}
                     url={linkContent}
-                    thumbnail={linkData?.thumbnails[0]}
+                    thumbnail={
+                      linkData?.thumbnails?.[0] ?? linkData?.icons?.[0]
```
ReidyT, line 217: Should we handle empty string here ?

```suggestion
                      linkData?.thumbnails?.[0] || linkData?.icons?.[0]
```

src/components/item/settings/ThumbnailSetting.tsx
```diff
@@ -96,10 +101,10 @@ const ThumbnailSetting = ({ item }: Props): JSX.Element | null => {
...
 
   const alt = translateBuilder(BUILDER.THUMBNAIL_SETTING_MY_THUMBNAIL_ALT);
-  const imgUrl = item.settings?.hasThumbnail
-    ? thumbnailUrl
-    : (item.type === ItemType.LINK ? item.extra[ItemType.LINK] : {})
-        ?.icons?.[0];
+  let imgUrl = item.settings?.hasThumbnail ? thumbnailUrl : null;
+  if (!imgUrl && item.type === ItemType.LINK) {
+    imgUrl = getLinkThumbnailUrl(item.extra);
+  }
```
ReidyT, line 107: Maybe you can use a function for that ?

```suggestion
const getImgUrl = (item, thumbnailUrl) => {
  if (item.settings?.hasThumbnail) {
    return thumbnailUrl;
  } else if (!imgUrl && item.type === ItemType.LINK) {
    return getLinkThumbnailUrl(item.extra);
  } else {
    return null;
  }
}

...

const imgUrl = getImgUrl();
```

src/components/item/settings/ThumbnailSetting.tsx
```diff
@@ -109,8 +114,9 @@ const ThumbnailSetting = ({ item }: Props): JSX.Element | null => {
           alt={alt}
           url={
             imgUrl ??
-            (item.type === ItemType.LINK ? item.extra[ItemType.LINK] : {})
-              ?.thumbnails?.[0]
+            (item.type === ItemType.LINK
+              ? getLinkThumbnailUrl(item.extra)
```
ReidyT, line 118: I'm not sure to understand here. If `imgUrl` is undefined, `getLinkThumbnailUrl(item.extra)` will return undefined no ? Because `imgUrl` will be set by `getLinkThumbnailUrl`, or am I missing something ?

src/components/table/ItemCard.tsx
```diff
@@ -77,13 +83,19 @@ const ItemCard = ({
     </Grid2>
   );
 
+  // show link thumbnail
+  let thumbnail = thumbnailUrl;
+  if (!thumbnail && item.type === ItemType.LINK) {
+    thumbnail = getLinkThumbnailUrl(item.extra);
+  }
+
```
ReidyT, line 91: You could reuse `getImgUrl()` if you decided to use a function in the previous component ;)

ReidyT (approved):
LGTM! I just let some suggestions 🚀

src/components/item/settings/ThumbnailSetting.tsx
pyphilia: I agree with your suggestion, I think anyway this code is a duplicata and should maybe be handled at the query client level (we always check thumbnail before we fallback to icon/thumbnail), I would even prefer save those thumbnails as normal thumbnail so we don't have problem checking for both each time. But it's future story..

PR #1350, opened 2024-07-16
feat: add member validation alert

### LINKED ISSUE #1339: Show a banner when member is not validated
With the changes in the backend where members need to be validated, we need to notify users in the frontend that they need to update their status.

spaenleh:
In this PR I add the member validation banner to alert users that their account needs to be validated.

Below is a screenshot.

Users have the possibility to close the banner, it will not be shown unless they reload the page.
Screenshot 2024-07-16 at 08 07 38

> WARNING (happens only on local dev)
> I have a small issue with the text not being updated when the language changes. Not sure if this is a change I did or if it was there before. 

fixes #1339


src/components/main/Main.tsx
```diff
@@ -47,12 +49,14 @@ export const platformsHostsMap = defaultHostsMapper({
...
 
 const Main = ({ children }: Props): JSX.Element => {
-  const { t } = useBuilderTranslation();
+  const { t, i18n } = useBuilderTranslation();
   const theme = useTheme();
   const { isMobile } = useMobileView();
 
   const itemId = useParams()[ITEM_ID_PARAMS];
 
+  useEffect(() => {}, [i18n.language]);
```
LinaYahya, line 58: I think it should be removed

LinaYahya (approved):

src/components/alerts/MemberValidationBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+};
+// eslint-disable-next-line arrow-body-style
+const buildLocalizedDocumentationLink = (lang: string): string => {
+  // english does not use a path prefix
+  return `${buildLocalizedDocumentationOrigin(lang)}${MEMBER_VALIDATION_DOCUMENTATION_LINK}`;
+};
+
+const MemberValidationBanner = (): JSX.Element | false => {
+  const [open, setOpen] = useState(true);
```
ReidyT, line 39: If you want, I made a really small hook that allow you to simplify the modal's state ^^. It's not a big improvements so it's really up to you ;)

```suggestion
  const { isOpen, openModal, closeModal } = useModalStatus({ isInitiallyOpen: true });
```

src/components/alerts/MemberValidationBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <Alert
+        id={MEMBER_VALIDATION_BANNER_ID}
+        severity="warning"
+        action={
+          <IconButton
+            id={MEMBER_VALIDATION_BANNER_CLOSE_BUTTON_ID}
+            onClick={() => {
+              setOpen(false);
+            }}
```
ReidyT, line 52:
```suggestion
            onClick={closeModal}
```

ReidyT (approved):
Nice PR, LGTM! 🚀

pyphilia (approved):
Thanks for the PR!

src/components/alerts/MemberValidationBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+const MEMBER_VALIDATION_DOCUMENTATION_LINK = '/user/account/validation';
+const buildLocalizedDocumentationOrigin = (lang: string = 'en') => {
+  switch (lang) {
+    case 'en':
+      return DOCUMENTATION_ORIGIN;
+    default:
+      return `${DOCUMENTATION_ORIGIN}/${lang}`;
+  }
+};
+// eslint-disable-next-line arrow-body-style
```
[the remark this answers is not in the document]
spaenleh: Yes because otherwise it complains about the fact that the function should not use return and be body-less

pyphilia (approved):

PR #1355, opened 2024-07-17
feat: move the publication status logic in backend

ReidyT:


cypress/support/server.ts
```diff
@@ -595,11 +596,7 @@ export const mockPostItemMembership = (
   cy.intercept(
     {
       method: HttpMethod.Post,
-      url: new RegExp(
-        `${API_HOST}/${parseStringToRegExp(
-          buildPostItemMembershipRoute(ID_FORMAT),
-        )}`,
-      ),
+      url: `${API_HOST}/item-memberships?*`,
```
spaenleh, line 599: I think you need to escape the `?` no ? 
```suggestion
      url: `${API_HOST}/item-memberships\?*`,
```

package.json
```diff
@@ -21,8 +21,8 @@
     "@emotion/styled": "11.11.5",
     "@graasp/chatbox": "3.1.0",
     "@graasp/map": "1.16.0",
-    "@graasp/query-client": "3.15.2",
-    "@graasp/sdk": "4.17.0",
+    "@graasp/query-client": "github:graasp/graasp-query-client#1115-determine-publication-state-backend",
+    "@graasp/sdk": "github:graasp/graasp-sdk#1115-determine-publication-state-backend",
```
spaenleh, line 25: Note to update

spaenleh (approved):
🚜 Looks good !

cypress/support/server.ts
ReidyT: No I don't think it is necessary:

`Unnecessary escape character: \?.eslint`

spaenleh: Okay good then !

PR #1359, opened 2024-07-18
feat: prevent changing visibility published item

ReidyT:
image


cypress/e2e/item/share/changeVisibility.cy.ts
```diff
@@ -125,4 +129,71 @@ describe('Visibility of an Item', () => {
       expect(url).to.include(item.id);
     });
   });
+
+  describe.only('Change visibility of published item', () => {
```
spaenleh, line 133: Remove `.only` ?

cypress/e2e/item/share/changeVisibility.cy.ts
```diff
@@ -125,4 +129,71 @@ describe('Visibility of an Item', () => {
...
+      const item = PublishedItemFactory(
+        PackedFolderItemFactory({}, { publicTag: {} }),
+      );
+      cy.setUpApi({
+        items: [item],
+        itemPublicationStatus: PublicationStatus.Published,
+      });
+      cy.visit(buildItemPath(item.id));
+      cy.get(`#${buildShareButtonId(item.id)}`).click();
+      const visiblitySelect = cy.get(
```
spaenleh, line 144:
```suggestion
      const visibilitySelect = cy.get(
```

spaenleh (approved):

cypress/e2e/item/share/changeVisibility.cy.ts
ReidyT: good catch ^^

PR #1367, opened 2024-07-19
feat: add unpublish button in outdated state

ReidyT:
* update ReadyToPublish state to be more like unpublished


pyphilia (commented):
Thanks for all the changes!
I'm actually wondering whether `validate` is still clear as a word on its own. Maybe we should be clearer like "Run verifications" (not good suggestion). Do you have any ideas? Or maybe a info icon to say which verifications we do. Plus it's not that clear that clicking on "validate" will actually publish the item. Maybe it should be rather "Validate & Publish"? Actually I think I'm okay with a button "publish" and have the status be "verification in progress" that will then become green once the validation is done.
I think it's a bit weird the "unpublish" is sometimes on the left or the right of the "contained" button. Maybe we should have the "unpublish" action be separated, like "destructive settings" of github?

Otherwise I'm all good with the code!

spaenleh (approved):
LGTM I also agree with the comments from Kim !

ReidyT:
After discussing with @pyphilia @spaenleh, the state outdate is not useful now because it doesn't check the children and nothing happen in the library. This state can be removed for now. We should think about a better way to handle updates (like revalidating at each change or freezing the published item and asking to publish the changes in the library).

PR #1383, opened 2024-07-25
feat: use backend search for accessible and children

pyphilia:


cypress/e2e/item/view/viewFolder.cy.ts
```diff
@@ -89,11 +89,25 @@ describe('View Folder', () => {
...
   it('search', () => {
     const { id } = parentItem;
+    const searchText = child1.name;
     cy.visit(buildItemPath(id, { mode: ItemLayoutMode.Grid }));
 
     cy.get(`#${buildItemCard(child1.id)}`).should('be.visible');
 
-    cy.get(`#${ITEM_SEARCH_INPUT_ID}`).type(child1.name);
+    cy.wait('@getChildren');
+    cy.wait('@getChildren');
```
ReidyT, line 98: Why do you need these two here ? Maybe add a comment if you really need to wait on them.
Also, I think the following should work too.

```suggestion
    cy.wait(['@getChildren', '@getChildren']);
```

cypress/e2e/item/view/viewFolder.cy.ts
```diff
@@ -89,11 +89,25 @@ describe('View Folder', () => {
...
+
+    cy.get(`#${ITEM_SEARCH_INPUT_ID}`).type(searchText);
+
+    cy.wait('@getChildren').then(({ request: { query } }) => {
```
ReidyT, line 102: It is possible to type the query to avoid the `as unknown` ? If not it's okay like that ^^

package.json
```diff
@@ -22,8 +22,8 @@
     "@emotion/styled": "11.13.0",
     "@graasp/chatbox": "3.1.0",
     "@graasp/map": "1.16.0",
-    "@graasp/query-client": "3.16.0",
-    "@graasp/sdk": "4.20.0",
+    "@graasp/query-client": "github:graasp/graasp-query-client#search-keywords",
```
ReidyT, line 25: just a reminder here ^^

ReidyT (approved):
Looks good ! I just let a few comments 😃

PR #1387, opened 2024-07-29
fix: prevent modal calls on selection

pyphilia:
Selecting items trigger lots of renders.
- Moved down DragSelection (so some duplication happens). Some manual adjustement is necessary because of margin and padding.
-  do not fetch items for modal if they are not opened.

### DIFF
--- a/src/components/item/FolderContent.tsx
+++ b/src/components/item/FolderContent.tsx
@@ -1,4 +1,4 @@
-import { Alert, Box, Stack, Typography } from '@mui/material';
+import { Alert, Box, Stack, Typography, useTheme } from '@mui/material';
 
 import {
   PackedItem,
@@ -27,6 +27,7 @@ import {
   SelectionContextProvider,
   useSelectionContext,
 } from '../main/list/SelectionContext';
+import { useDragSelection } from '../main/list/useDragSelection';
 import { DesktopMap } from '../map/DesktopMap';
 import NoItemFilters from '../pages/NoItemFilters';
 import SortingSelect from '../table/SortingSelect';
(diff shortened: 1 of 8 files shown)


src/components/item/FolderContent.tsx
```diff
@@ -49,6 +50,10 @@ const Content = ({ item, searchText, items, sortBy }: Props) => {
   const { itemTypes } = useFilterItemsContext();
   const { selectedIds, clearSelection, toggleSelection } =
     useSelectionContext();
+  const theme = useTheme();
+  const DragSelection = useDragSelection({
+    adjustments: { marginTop: 70, marginLeft: theme.spacing(3) },
```
spaenleh, line 55: Where are these defined ? can we use the same constants ?

src/components/main/list/useDragSelection.tsx
```diff
@@ -0,0 +1,76 @@
...
+    onSelectionEnd: () => {},
+    selectionProps: {
+      style: {
+        // adjustement
+        // https://github.com/AirLabsTeam/react-drag-to-select/issues/30
+        ...adjustments,
+        border: `2px dashed ${PRIMARY_COLOR}`,
+        borderRadius: 4,
+        backgroundColor: 'lightblue',
+        opacity: 0.5,
```
spaenleh, line 69: Should we also increase the zIndex so it appears on top of the drawer ?

spaenleh (approved):

src/components/pages/RecycledItemsScreen.tsx
```diff
@@ -51,70 +53,77 @@ const RecycledItemsScreenContent = ({
...
-            </Stack>
-          )}
+            )}
+          </Stack>
+          {
+            // render the filtered data and when it is empty display that nothing matches the search
+            filteredData?.length ? (
+              filteredData.map((item) => (
+                <ItemCard
+                  // todo: should not be able to click on the card
```
ReidyT, line 102: Still something to do ?

ReidyT (approved):
LGTM 👑

src/components/item/FolderContent.tsx
pyphilia: These depends on the layout.. so it's kind of unstable, but it depends on many params unfortunately =/

PR #1388, opened 2024-07-29
fix: prevent map to loop

pyphilia:


spaenleh (approved):
Interesting, can you quickly explain what was the cause of the issue for the loop ? 

GG on finding the cause btw ! 💪

ReidyT (approved):
Wow good catch ! Yeah, I'm also interested to know why this caused the issue 😆

pyphilia:
Honestly I'm not sure, I corrected it because it was not really correct. I guess this works currently because we don't change language in the same platform, so no need to listen on i18n, but I'm not sure why the map was triggering looping :/

PR #1396, opened 2024-08-05
fix: does not trigger drag on mousedown svg

pyphilia:
Currently there's a small bug happening if you drag and you specifically click on an icon (svg), it will allow you to drag but then it will trigger the multi selection. 

This should fix the problem.


ReidyT (approved):
LGTM! Nice fix

Issue #1431, opened 2024-09-11
Use only lucide icons


spaenleh:
- [x] Chat icon is stroke through 
- [x] Library icon is not a lucide one
- [x] Pin icon is strike through
- [x] In the trash restore is not a lucide icon
- [x] Hide icons is not lucide
- [x] create shortcut icon is not lucide 
- [x] duplicate icon is not lucide

Goal of the change:
<img width="2239" alt="Screenshot 2024-09-11 at 14 59 09" src="https://github.com/user-attachments/assets/73232239-002a-48b0-ab1e-964fa51f8f58">

PR #1432, opened 2024-09-11
fix: udpate menu icons to lucide

spaenleh:
In this PR I updated the icons of the main menu to use lucide alternatives, below is a before and after:

Screenshot 2024-09-11 at 11 37 30
Screenshot 2024-09-11 at 11 37 22

ref #1431

### DIFF
--- a/src/components/main/Main.tsx
+++ b/src/components/main/Main.tsx
@@ -14,6 +14,7 @@ import {
 
 import { HOST_MAP } from '@/config/externalPaths';
 import { useBuilderTranslation } from '@/config/i18n';
+import { hooks } from '@/config/queryClient';
 import { BUILDER } from '@/langs/constants';
 
 import { HOME_PATH, ITEM_ID_PARAMS } from '../../config/paths';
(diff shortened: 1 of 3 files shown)


src/components/main/Main.tsx
```diff
@@ -82,6 +84,7 @@ const Main = ({ children }: Props): JSX.Element => {
   );
   return (
     <GraaspMain
+      open={currentMember ? undefined : false}
```
ReidyT, line 87: Why it is undefined instead of true? It is not the same to give false or undefined?

ReidyT (commented):
Thanks for this PR! I'm just not sure to understand why the `GraaspMain`'s open property take false or undefined.

ReidyT (approved):

src/components/main/Main.tsx
spaenleh: As said in https://github.com/graasp/graasp-builder/pull/1434#discussion_r1754318118 I want to only set the prop when the user is **not logged in**, in the other cases I want to keep the default behaviour.

PR #1434, opened 2024-09-11
fix: update mode icon

spaenleh:
Based on #1432 
- **fix: update mode icons**

Screenshot 2024-09-11 at 11 54 30
Screenshot 2024-09-11 at 11 56 08

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,9 @@
     "@graasp/map": "1.17.0",
     "@graasp/query-client": "3.22.3",
     "@graasp/sdk": "4.26.0",
+    "@graasp/stylis-plugin-rtl": "2.2.0",
     "@graasp/translations": "1.35.1",
-    "@graasp/ui": "4.26.0",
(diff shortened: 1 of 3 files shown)


src/components/main/Main.tsx
```diff
@@ -82,6 +84,7 @@ const Main = ({ children }: Props): JSX.Element => {
   );
   return (
     <GraaspMain
+      open={currentMember ? undefined : false}
```
pyphilia, line 87:
```suggestion
      open={Boolean(currentMember)}
```

pyphilia (approved):

src/components/main/Main.tsx
spaenleh: I want to use `undefined` so the default behaviour can be kept, i.e. when on mobile it will be closed, otherwise mobile users that are logged in will always have it open.

ReidyT (approved):

PR #1435, opened 2024-09-11
fix: update item menu icons

spaenleh:
based on #1434 
- **fix: menu icons**


src/components/common/DeleteButton.tsx
```diff
@@ -46,7 +46,10 @@ const DeleteButton = ({
     onClose?.();
   };
 
-  const text = translateBuilder(BUILDER.DELETE_BUTTON);
+  const text =
+    items.length > 1
+      ? translateBuilder(BUILDER.DELETE_ALL_SELECTED_ITEMS_BUTTON)
+      : translateBuilder(BUILDER.DELETE_BUTTON);
```
ReidyT, line 52: I think you could use the plurals of i18next here (https://www.i18next.com/translation-function/plurals)

src/components/common/RestoreButton.tsx
```diff
@@ -28,7 +29,11 @@ const RestoreButton = ({
     onClickFn?.();
   };
 
-  const title = translateBuilder(BUILDER.RESTORE_ITEM_BUTTON);
+  const title =
+    // adapt the tooltip to the number of items that will be restored
+    itemIds.length > 1
+      ? translateBuilder(BUILDER.RESTORE_ALL_SELECTED_ITEMS_BUTTON)
+      : translateBuilder(BUILDER.RESTORE_ITEM_BUTTON);
```
ReidyT, line 36: Same comment here

src/components/main/ItemMenuContent.tsx
ReidyT, line 1: I really like the idea of handling permissions, but I'm wondering if it's more readable than before? Maybe we could try to improve it a bit (it could be done in another PR).

ReidyT (approved):
Thanks for this PR, LGTM! I just let two comments 😄

src/components/main/ItemMenuContent.tsx
spaenleh: I see, my main goal was to be able to add the dividers programatically between each section, that could have a different number of elements, and sometimes be completely empty. I agree this does not lead to the most clean solution.. But I hope it is a bit more robust than before. I am open to discuss improvements with a generic function to do the work of interspersing the dividers etc ...

src/components/item/header/Actions.tsx
```diff
@@ -104,20 +104,16 @@ const Actions = ({ item }: Props): JSX.Element | null => {
...
+        <Divider />
         <FlagButton item={item} />
+        {canAdmin && (
+          <RecycleButton
+            key="recycle"
+            type={ActionButton.MENU_ITEM}
+            itemIds={[item.id]}
+            onClick={closeMenu}
+          />
+        )}
```
pyphilia, line 116: We do more often a delete than a flag, shouldn't this be higher in the menu then?

pyphilia (approved):
LGTM 🎨

src/components/main/ItemMenuContent.tsx
pyphilia: If we have this magic permission table, perhaps we could just generate a map, or a true/false function that show or not the element. Not sure but it's sure not so easy to handle. Or a different component per permission?

ReidyT: I understand and I agree that it's not easy, I don't have a solution now (I haven't really think about it). I think it's really good enough, and we can investigate more later :)

PR #1460, opened 2024-09-18
feat: show item login message on customer pages

pyphilia:
https://github.com/user-attachments/assets/c7daf799-25f0-4dc4-98c1-9cff97cbfc95


ReidyT (approved):
Thanks for this PR! LGTM 🚀

src/components/App.tsx
```diff
@@ -84,26 +67,73 @@ const App = (): JSX.Element => {
...
             <Route path={ITEM_PUBLISH_PATH} element={<LibrarySettingsPage />} />
             <Route path={ITEM_SETTINGS_PATH} element={<ItemSettingsPage />} />
           </Route>
         </Route>
-        <Route path={RECYCLE_BIN_PATH} element={<RecycleWithAuthorization />} />
-        <Route path={buildItemPath()} element={<ItemScreen />} />
-        <Route path={ITEMS_PATH} element={<HomeWithAuthorization />} />
+
+        {/* redirection to home */}
+        <Route path={ITEMS_PATH} element={<Redirect />} />
```
spaenleh, line 136: What is this item path ?

src/components/item/header/Actions.tsx
```diff
@@ -77,34 +79,47 @@ const Actions = ({ item }: Props): JSX.Element | null => {
...
+                openCreateShortcutModal();
+                closeMenu();
+              }}
+            />
+            <BookmarkButton
+              size="medium"
+              key="bookmark"
+              type={ActionButton.MENU_ITEM}
+              item={item}
+              className={ITEM_MENU_BOOKMARK_BUTTON_CLASS}
```
spaenleh, line 96: Selection using className ? Can we use an id instead ? I think we should move away from classes for selectors.

src/components/pages/home/HomeScreen.tsx
```diff
@@ -197,28 +209,78 @@ const HomeScreenContent = ({ searchText }: { searchText: string }) => {
...
+                  startIcon={<ClipboardPen />}
+                  variant="contained"
+                  sx={{ textTransform: 'none' }}
+                >
+                  {translateBuilder(BUILDER.GUEST_SIGN_OUT_BUTTON)}
+                </MuiButton>
+              </Box>
+            </Alert>
+          </Container>
         </Stack>
```
spaenleh, line 244: This looks a lot like the `PreventGuest` component ?

src/langs/fr.json
```diff
@@ -71,7 +71,7 @@
   "EDIT_ITEM_ERROR_MESSAGE": "L'élément n'est pas valide",
   "EDIT_ITEM_MODAL_TITLE": "Modifier l'élément",
   "EDIT_ITEM_IMAGE_ALT_TEXT_LABEL": "Texte alternatif (pour améliorer l'accessibilité)",
-  "EMPTY_ITEM_MESSAGE": "Cet élément est vide.",
+  "EMPTY_FOLDER_MESSAGE": "Ce folder est vide.",
```
spaenleh, line 74:
```suggestion
  "EMPTY_FOLDER_MESSAGE": "Ce dossier est vide.",
```

spaenleh (changes requested):
Thank you for the work. 💪 

It looks like you are fixing multiple issues in the same PR. 

I am not sure why we are doing some verifications on the HomeScreen component when we should already have them embedded with the wrappers defined in App.tsx ?

src/components/App.tsx
pyphilia: I guess it's /items that does not have any id, so you return to home.

spaenleh: This will be handled by the "catchall" route defined at the end no ?

src/components/item/header/Actions.tsx
pyphilia: We sometimes use classnames because `id` doesn't work. Here I think it's because the graasp-ui component does not allow me to set an id, but allows a class names.
Maybe we should resolve this issue alttogether from ui.

spaenleh (approved):

PR #1499, opened 2024-10-03
fix: fix edit modal content and navigation

pyphilia:
This PR fixes two problems:

- navigating from item root to child was not updating the parent. We removed the opti `path` that was unsync with `itemId`
- we fixed the edit item modal content, by adding a `useEffect`. This is just some tape on the problem for now, we should refactor the full component. #1498

### DIFF
--- a/cypress/e2e/item/edit/editFolder.cy.ts
+++ b/cypress/e2e/item/edit/editFolder.cy.ts
@@ -3,7 +3,10 @@ import { PackedFolderItemFactory } from '@graasp/sdk';
 import { HOME_PATH, buildItemPath } from '../../../../src/config/paths';
 import {
   EDIT_ITEM_BUTTON_CLASS,
+  EDIT_ITEM_MODAL_CANCEL_BUTTON_ID,
+  FOLDER_FORM_DESCRIPTION_ID,
   ITEM_FORM_CONFIRM_BUTTON_ID,
+  ITEM_FORM_NAME_INPUT_ID,
   buildItemsGridMoreButtonSelector,
 } from '../../../../src/config/selectors';
 import { EDIT_ITEM_PAUSE } from '../../../support/constants';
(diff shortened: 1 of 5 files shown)


cypress/support/server.ts
```diff
@@ -1012,15 +1012,8 @@ export const mockGetItemLoginSchemaType = (items: ItemForTest[]): void => {
...
-      const type = item?.itemLoginSchema?.type;
-      if (!type) {
-        return reply({
-          statusCode: StatusCodes.NOT_FOUND,
-        });
-      }
-
       return reply({
-        body: type,
+        body: item?.itemLoginSchema?.type ?? null,
```
pyphilia, line 1016: This is the actual response of the backend.

cypress/e2e/item/edit/editFolder.cy.ts
```diff
@@ -94,4 +97,30 @@ describe('Edit Folder', () => {
       },
     );
   });
+
+  // moving from parent to child might not update info since the item folder component is the same
```
ReidyT, line 101:
```suggestion
  // navigating from parent to child might not update info since the item folder component is the same
```

cypress/e2e/item/edit/editFolder.cy.ts
```diff
@@ -94,4 +97,30 @@ describe('Edit Folder', () => {
...
+  it('edit 2 folders should display the correct data', () => {
+    const parentItem = PackedFolderItemFactory();
+    const itemToEdit = PackedFolderItemFactory({
+      parentItem,
+      description: 'first description',
+    });
+    cy.setUpApi({ items: [parentItem, itemToEdit] });
+    // go to children item
```
ReidyT, line 109:
```suggestion
    // go to parent item
```

cypress/support/server.ts
ReidyT, line 1016: Should we add a comment directly in the code?

ReidyT (approved):
LGTM ⛳️

PR #1500, opened 2024-10-03
feat: add umami tags to track actions

### LINKED ISSUE #1495: Setup Umami for tracking

spaenleh:
In this PR: 
- [x] add umami script
- [x] add umami events on the home page

Still todo is #1502 

Umami-events (1)

closes #1495

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,12 +20,15 @@
     />
 
     <!-- Umami tracking code -->
-    <!-- TODO: use an env var -->
+
+    <!--
+    The integrity can not be checked when using cross origin
+    integrity="sha384-JyQQf9XxkMfVx/h+OQ+SxUZqLbL3p3KmWfnZ4LpQWzksmzHXESgXCk1/5AYxT9nR"
+    -->
     <script
       defer
       src="https://cloud.umami.is/script.js"
-      data-website-id="46068c2d-17a3-4fe2-a1f0-1a0e992711f8"
-      integrity="sha384-JyQQf9XxkMfVx/h+OQ+SxUZqLbL3p3KmWfnZ4LpQWzksmzHXESgXCk1/5AYxT9nR"
+      data-website-id="%VITE_UMAMI_WEBSITE_ID%"
     ></script>
     <title>Graasp Builder</title>
   </head>
--- a/index.html
+++ b/index.html
@@ -18,6 +18,14 @@
       rel="stylesheet"
       href="https://fonts.googleapis.com/css2?family=Nunito:ital,wght@0,200..1000;1,200..1000&display=swap"
     />
+
+    <!-- Umami tracking code -->
+    <!-- TODO: use an env var -->
+    <script
+      defer
+      src="https://cloud.umami.is/script.js"
+      data-website-id="46068c2d-17a3-4fe2-a1f0-1a0e992711f8"
+    ></script>
     <title>Graasp Builder</title>
   </head>
   <body>
--- a/index.html
+++ b/index.html
@@ -25,6 +25,7 @@
       defer
       src="https://cloud.umami.is/script.js"
       data-website-id="46068c2d-17a3-4fe2-a1f0-1a0e992711f8"
+      integrity="sha384-JyQQf9XxkMfVx/h+OQ+SxUZqLbL3p3KmWfnZ4LpQWzksmzHXESgXCk1/5AYxT9nR"
     ></script>
     <title>Graasp Builder</title>
   </head>
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@graasp/sdk": "4.31.0",
     "@graasp/stylis-plugin-rtl": "2.2.0",
     "@graasp/translations": "1.39.0",
-    "@graasp/ui": "5.2.2",
+    "@graasp/ui": "5.3.0",
     "@mui/icons-material": "5.16.7",
     "@mui/lab": "5.0.0-alpha.173",
     "@mui/material": "5.16.7",
--- a/src/components/common/SelectTypes.tsx
+++ b/src/components/common/SelectTypes.tsx
@@ -68,7 +68,12 @@ export const SelectTypes = (): JSX.Element => {
         multiple
         value={itemTypes}
         onChange={handleChange}
-        input={<OutlinedInput label={label} />}
+        input={
+          <OutlinedInput
+            label={label}
+            inputProps={{ 'data-umami-event': 'filter-item-type' }}
+          />
+        }
         renderValue={renderValues}
         MenuProps={MenuProps}
         sx={{
--- a/src/components/common/UserSwitchWrapper.tsx
+++ b/src/components/common/UserSwitchWrapper.tsx
@@ -49,7 +49,7 @@ const UserSwitchWrapper = ({ ButtonContent }: Props): JSX.Element => {
       seeProfileButtonId={HEADER_MEMBER_MENU_SEE_PROFILE_BUTTON_ID}
       buildMemberMenuItemId={buildMemberMenuItemId}
       avatar={<MemberAvatar id={member?.id} />}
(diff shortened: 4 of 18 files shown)


index.html
```diff
@@ -18,6 +18,15 @@
       rel="stylesheet"
       href="https://fonts.googleapis.com/css2?family=Nunito:ital,wght@0,200..1000;1,200..1000&display=swap"
     />
+
+    <!-- Umami tracking code -->
+    <!-- TODO: use an env var -->
```
pyphilia, line 23: Todo now?

package.json
```diff
@@ -24,7 +24,7 @@
     "@graasp/sdk": "4.31.0",
     "@graasp/stylis-plugin-rtl": "2.2.0",
     "@graasp/translations": "1.39.0",
-    "@graasp/ui": "5.2.0",
+    "@graasp/ui": "github:graasp/graasp-ui#main-umami",
```
pyphilia, line 27: Reminder!

src/components/main/Main.tsx
```diff
@@ -35,8 +36,14 @@ const StyledLink = styled(Link)(() => ({
   display: 'flex',
   alignItems: 'center',
 }));
-const LinkComponent = ({ children }: { children: JSX.Element }) => (
-  <StyledLink to={HOME_PATH}>{children}</StyledLink>
+const LinkComponent = ({ children }: { children: ReactNode }) => (
+  <StyledLink
+    data-umami-event="header-home-link"
+    data-umami-event-context={Context.Builder}
```
pyphilia, line 42: Maybe we want `event-page` as well?

src/components/main/MainMenu.tsx
```diff
@@ -36,7 +36,11 @@ const ResourceLinks = () => {
   const { t } = useBuilderTranslation();
   return (
     <ListItem disablePadding>
-      <ListItemButton href={TUTORIALS_LINK} target="_blank">
+      <ListItemButton
+        href={TUTORIALS_LINK}
+        target="_blank"
+        data-umami-event="sidebar-tutorials"
```
pyphilia, line 42: Have `event-page`? Or will it  result in a different chart?

src/components/common/SelectTypes.tsx
```diff
@@ -68,7 +68,12 @@ export const SelectTypes = (): JSX.Element => {
         multiple
         value={itemTypes}
         onChange={handleChange}
-        input={<OutlinedInput label={label} />}
+        input={
+          <OutlinedInput
+            label={label}
+            inputProps={{ 'data-umami-event': 'filter-item-type' }}
```
pyphilia, line 74: Add event-page to know where?

src/components/main/NewItemButton.tsx
```diff
@@ -45,6 +45,7 @@ const NewItemButton = ({
             background: theme.palette.primary.main,
             '&:hover': { background: 'grey' },
           }}
+          data-umami-event="new-item-icon-button"
```
pyphilia, line 48: Add page? or context? could be add in home or in children.

src/components/main/NewItemButton.tsx
```diff
@@ -56,6 +57,7 @@ const NewItemButton = ({
           aria-label="add"
           startIcon={<PlusIcon />}
           size={size}
+          data-umami-event="new-item-button"
```
pyphilia, line 60: Add page? or context? could be add in home or in children.

src/components/main/Main.tsx
pyphilia: I wonder if we should create some kind of type, and spread. You could have the same type in ui if you need to pass the params. Maybe overengineered?

```
type Type = {
   event: string;
  eventContext?: string
}

const buildUmamiProps = (args: Type) => ({ 
  data-umami-event: args.event, 
  data-umami-event-context: args.eventContext
})
```

src/components/main/Main.tsx
```diff
@@ -35,8 +36,14 @@ const StyledLink = styled(Link)(() => ({
   display: 'flex',
   alignItems: 'center',
 }));
-const LinkComponent = ({ children }: { children: JSX.Element }) => (
-  <StyledLink to={HOME_PATH}>{children}</StyledLink>
+const LinkComponent = ({ children }: { children: ReactNode }) => (
+  <StyledLink
+    data-umami-event="header-home-link"
```
pyphilia, line 41: I wonder if it's okay to send `header.home-link`? so the scope is clearer?

pyphilia (changes requested):
Thanks for this change! Let me know what you think of my suggestion.
I think `event-page` should be always returned?

index.html
spaenleh: Done. I think it can be done better, by for example defining a custom vite plugin that could insert the tag if the variable is defined. Here is an example: https://github.com/KermanX/vite-plugin-ms-clarity/blob/main/plugin.ts

src/components/main/Main.tsx
spaenleh: The page is already send with each event. The UI is however not really designed to show a specific graph for events by page. But you can filter for a specific page and then you will see the exact number of invocations for each event on that page specifically. I don't know if this is enough. Maybe for the moment ?

src/components/main/NewItemButton.tsx
spaenleh: Do we really need to know if it was in children or in home ?

src/components/main/Main.tsx
spaenleh: I think the event names should be consistent, not sure we need to add complexity to them. We can discuss the naming scheme

index.html
```diff
@@ -18,6 +18,18 @@
       rel="stylesheet"
       href="https://fonts.googleapis.com/css2?family=Nunito:ital,wght@0,200..1000;1,200..1000&display=swap"
     />
+
+    <!-- Umami tracking code -->
+
+    <!--
+    The integrity can not be checked when using cross origin
+    integrity="sha384-JyQQf9XxkMfVx/h+OQ+SxUZqLbL3p3KmWfnZ4LpQWzksmzHXESgXCk1/5AYxT9nR"
+    -->
```
pyphilia, line 27: Remove?

src/components/common/UserSwitchWrapper.tsx
```diff
@@ -49,6 +49,7 @@ const UserSwitchWrapper = ({ ButtonContent }: Props): JSX.Element => {
       seeProfileButtonId={HEADER_MEMBER_MENU_SEE_PROFILE_BUTTON_ID}
       buildMemberMenuItemId={buildMemberMenuItemId}
       avatar={<MemberAvatar id={member?.id} />}
+      dataUmami="header-avatar"
```
pyphilia, line 52:
```suggestion
      dataUmamiEvent="header-avatar"
```

src/components/item/ItemSearch.tsx
```diff
@@ -51,6 +51,7 @@ export const useItemSearch = ({
       inputBaseId={ITEM_SEARCH_INPUT_ID}
       placeholder={translateBuilder(BUILDER.ITEM_SEARCH_PLACEHOLDER)}
       size="small"
+      dataUmami="item-search"
```
pyphilia, line 54:
```suggestion
      dataUmamiEvent="item-search"
```

pyphilia (approved):
All good for me, I would use `dataUmamiEvent` it's more precise.

index.html
spaenleh: So, this integrity property was recommended by Sonarcloud, to ensure that the code that is downloaded actually meets what is expected, so it increases security because you can not download code that would have changed without you knowing it. 

The issue is that it does not work for cross-origin requests which is our case currently. It could work if we host umami ourselves AND we move to a single origin. So I can keep it in a comment, but yeah could also remove it.

PR #1514, opened 2024-10-10
fix: update crop modal to not inflate image sizes

spaenleh:
In this PR I refactor the crop modal execution to fix an issue where the thumbnail would be artificially inflated, resulting in very large uploads. 

This code was taken from account where this fix was initially made.


ReidyT (approved):
Nice fix, LGTM

PR #1523, opened 2024-10-14
refactor: split document form into create and edit forms

pyphilia:
Split `DocumentForm` into `CreateDocumentForm` and `EditDocumentForm`, and added some smaller components.
Partially moved to `react-hook-form`. `TextEditor` being a bit more tricky, I've hacked around it.

We still use global `setChanges` for compatibility, as long as all other forms aren't refactored, and that as long as it's the modal that contains the `<form>`.

ref #1498 #1522


spaenleh (approved):
Looks good, thank you for your work !

PR #1529, opened 2024-10-18
refactor: use react hook form for base item form

pyphilia:
ref #1522


spaenleh (approved):
It feels very weird to use react-hook-form like this. This is only a transition state right ? Otherwise if we plan on keeping nested structures we should invest in using other methods from react-hook-form like the contexts and useController hook.

pyphilia:
Yes it's a transitional state, in the end we should remove the `useEffect` and `watch`.

PR #1550, opened 2024-10-25
feat: allow press enter to share item

pyphilia:
Use react-hook-form for Sharing item. It allows to press enter to validate form.
Fixed cancel button as well.


src/components/item/sharing/shareButton/CreateItemMembershipForm.tsx
```diff
@@ -134,59 +116,61 @@ const CreateItemMembershipForm = ({
...
+            spacing={1}
+          >
+            <TextField
+              id={SHARE_ITEM_EMAIL_INPUT_ID}
+              variant="outlined"
+              label={translateBuilder(BUILDER.SHARE_ITEM_FORM_EMAIL_LABEL)}
+              helperText={errors.email?.message}
+              {...register('email', {
+                required: true,
+                validate: checkForInvitationError,
```
spaenleh, line 145: You can pass multiple validation criteria:
```suggestion
                validate: { isEmail: (v) => validator.isEmail(v) || 'SOME_STRING', alreadyExists: checkForInvitationError},
```
It would allow to split up the validation function that you have currently.

src/components/item/sharing/shareButton/CreateItemMembershipForm.tsx
```diff
@@ -91,41 +90,24 @@ const CreateItemMembershipForm = ({
...
           },
         ],
       });
 
       // reset email input
-      setInvitation({
-        ...invitation,
-        email: '',
-      });
+      reset({ email: '', permission: PermissionLevel.Read });
```
spaenleh, line 110: I think you would not need to reset the form if the dialog is not persisted. You can try to instantiate the Dialog Content in a different component than the `Dialog` wrapper, so the state that controls the dialog is not inside the dialog content and the dialog can be destroyed and your form should be reset correctly.

spaenleh (changes requested):
Nice ! 

Can you add a test that cancels the dialog ?

cypress/e2e/memberships/createItemMembership.cy.ts
```diff
@@ -69,7 +73,71 @@ describe('Create Membership', () => {
...
+
+    // share
+    const member = MEMBERS.FANNY;
+    const permission = PermissionLevel.Admin;
+    shareItem({
+      id,
+      email: `${member.email}{Enter}`,
+      permission,
+      submit: false,
+    });
```
ReidyT, line 121: I'm wondering if we shouldn't press the enter after calling share item? In this way, it's clearer to see where and when the enter is done. What do you think?

ReidyT (approved):
LGTM 🤖

spaenleh (approved):

Issue #1551, opened 2024-10-25
Remove `updatedProperties` and `setChanges`, fully move to react-hook-form


pyphilia:

pyphilia:
All modals have been moved.

PR #1552, opened 2024-10-25
feat: submit individually file data for edit

### LINKED ISSUE #1522: Split Edit and Create forms, partially move to react-hook-form

pyphilia:
- Create ShortcutEditForm and apply react-hook-form
- Modify LinkForm (edit) so it fully uses react-hook-form. It sends itself the `editItem` request, so we avoid drilling up props.
  - also fix the bug of editing files by sending only `altText` for editing `extra` 

Once all forms are refactored, next move would be to use specific endpoints for each item type. 

ref #1551 
close #1522

### DIFF
--- a/src/components/item/edit/EditModal.tsx
+++ b/src/components/item/edit/EditModal.tsx
@@ -30,15 +30,14 @@ import { BUILDER } from '../../../langs/constants';
 import BaseItemForm from '../form/BaseItemForm';
 import FileForm from '../form/FileForm';
 import FolderForm from '../form/FolderForm';
-import NameForm from '../form/NameForm';
 import DocumentForm from '../form/document/DocumentForm';
+import EditShortcutForm from '../shortcut/EditShortcutForm';
 
 const { editItemRoutine } = routines;
 
 export interface EditModalContentPropType {
   item?: DiscriminatedItem;
   setChanges: (payload: Partial<DiscriminatedItem>) => void;
-  updatedProperties: Partial<DiscriminatedItem>;
 }
 export type EditModalContentType = CT<EditModalContentPropType>;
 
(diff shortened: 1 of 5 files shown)


src/components/item/form/FileForm.tsx
```diff
@@ -32,60 +43,94 @@ type Inputs = {
...
+      }
+      if (item.type === ItemType.LOCAL_FILE) {
+        return {
+          [ItemType.LOCAL_FILE]: {
+            altText,
+          },
+        } as LocalFileItemExtra;
+      }
+    }
+    console.error('item type is not handled');
```
ReidyT, line 86:
```suggestion
    console.error(`item type ${item.type} is not handled`);
```

src/components/item/edit/EditModal.tsx
```diff
@@ -131,6 +123,48 @@ const EditModal = ({ item, onClose, open }: Props): JSX.Element => {
...
+        </DialogContent>
+        <DialogActions>
+          <CancelButton
+            id={EDIT_ITEM_MODAL_CANCEL_BUTTON_ID}
+            onClick={onClose}
+          />
+          <Button
+            // should not allow users to save if the item is not valid
+            disabled={
+              // // maybe we do not need the state variable and can just check the item
```
ReidyT, line 150:
```suggestion
              // maybe we do not need the state variable and can just check the item
```

ReidyT: Is this comment still valid?

src/components/item/edit/EditModal.tsx
```diff
@@ -131,6 +123,48 @@ const EditModal = ({ item, onClose, open }: Props): JSX.Element => {
...
+              // isConfirmButtonDisabled ||
+              // isItem Valid checks a full item, so we add the updated properties to the item to check
```
ReidyT, line 152:
```suggestion
              // isItemValid checks a full item, so we add the updated properties to the item to check
```

ReidyT (approved):
LGTM 🪐

src/components/item/edit/EditModal.tsx
pyphilia: probably not, but this code will disappear anyway, I'll let it be for now.

PR #1580, opened 2024-11-22
refactor: use react hook form for document forms

pyphilia:
Split Document form into create and edit.
Finally could remove some dirty code in NewItemModal.

ref https://github.com/graasp/graasp-builder/issues/1551


ReidyT (approved):
LGTM 🌝

PR #1581, opened 2024-11-22
refactor: refactor h5p form

pyphilia:
ref #1551

### DIFF
--- a/src/components/main/ImportH5P.tsx
+++ b/src/components/main/ImportH5P.tsx
@@ -1,3 +1,4 @@
+import { ChangeEventHandler } from 'react';
 import { useParams } from 'react-router-dom';
 
 import {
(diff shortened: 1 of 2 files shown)


src/components/main/ImportH5P.tsx
```diff
@@ -1,72 +1,97 @@
...
+                  previousItemId,
+                  file: e.target.files[0],
+                })
+                  .then(() => {
+                    closeNotification();
+                    onClose?.();
+                  })
+                  .catch((error) => {
+                    closeNotification(error);
+                  });
```
ReidyT, line 76:
```suggestion
                try {
                  await importH5P({
                    onUploadProgress: update,
                    id: itemId,
                    previousItemId,
                    file: e.target.files[0],
                  });
                  
                  closeNotification();
                  onClose?.();
                } catch (error) {
                    closeNotification(error);
                };
```

We could also handle this logic in a function?

ReidyT (approved):
LGTM 🪐

PR #1582, opened 2024-11-22
refactor: refactor edit base item form

pyphilia:
I updated the base item form (edit modal for link, app, etherpad and h5p).
I removed the description placement in the settings page, it was not matching well with my refactor, and it didn't make much sense to have it in the settings. We can discuss it. 

ref #1551


ReidyT (approved):
Nice 😆

PR #1591, opened 2024-12-09
fix: link crash when url is not valid

### LINKED ISSUE #1590: Interface Crash on link creation
The interface crashes on link creation if the url is not fully qualified.

spaenleh:
fix #1590


ReidyT (approved):
LGTM 👍
Is the Cypress test working?

spaenleh:
@ReidyT I was not able to make the cypress test fail with the previous implementation ... so it is still passing now, even if I changed the code. Not sure how we should go about that ...

